PR #20, opened 2026-03-03
Fix/anthropic api

aniemela-amd:
Implements the Anthropic model in src/models/anthropic_model.py that was left as a stub. Enables using Anthropic API directly without AMD LLM Gateway.

Code is largely reused from the AMD LLM Gateway implementation in src/models/amd_claude.py.


src/minisweagent/models/anthropic_model.py
```diff
@@ -2,36 +2,286 @@
 
 import os
 import warnings
-from typing import Literal
+from dataclasses import asdict, dataclass, field
+from typing import Any, Literal
```
peyron-amd:
```suggestion
import logging
```

PR #59, opened 2026-03-23
Af/heterogeneous postprocess

sdubagun-amd:
Support for heterogenous mode and post-processing.

### DIFF
--- a/src/minisweagent/run/postprocess/evaluation.py
+++ b/src/minisweagent/run/postprocess/evaluation.py
@@ -49,9 +49,7 @@ class PatchApplyError(Exception):
     pass
 
 
-def _find_agent_worktree_slot(
-    results_dir: Path, best_task: str, repo_root: str, output_dir: Path
-) -> Path | None:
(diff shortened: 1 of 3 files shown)


src/minisweagent/run/dispatch.py
yueliu14: What module dose the dispatch belongs to? Please move to target directory

entrypoint.sh
yueliu14: Is entrypoint a necessary part? I think the README.md will tell users how to use GEAK v3.

jianghui-jianghui:
why N/A in the benchmark table? @iraj465

src/minisweagent/run/orchestrator.py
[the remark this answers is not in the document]
sdubagun-amd: pipeline_types have the dataclass contracts we introduced between different modules.

entrypoint.sh
sdubagun-amd: Entrypoint is a necessary script for Dockerfile to work correctly.

PR #91, opened 2026-03-31
add hip/triton examples for users test

yueliu14:
1. add hip/triton examples for users test
2. change README.md
3. auto infer num_parallel from length of gpu-ids


sdubagun-amd (changes requested):

jianghui-jianghui (approved):

sdubagun-amd (approved):
This is fine. One optional suggestion: `kernel_path` is already a supported alias to the `kernel_url` argument currently; see [mini.py#L179](https://github.com/AMD-AGI/GEAK/blob/6327319b2f3b5120b704bc12b1efabca049dfd81/src/minisweagent/run/mini.py#L179), so the instruction change regarding the argument need not be strict.

PR #120, opened 2026-04-09
fix: apply starting_patch before candidate in eval worktree (#110)

### LINKED ISSUE #110: `git apply patch` doesn't work with multiple rounds
### Describe the bug

since patches are always relative to the original kernel codes, the base code needs to be in the original version to be applied. For round3, the base code is from round2, which is the round1 code applied with the best generated patch, so the git apply will fail. So for the rounds after round3, no performance gain was observed.

Image

iraj465:
Multi-round patch application failed in two places:

1. evaluation.py (eval worktree): Round N+1 candidate patches are relative to the round N patched base, not the original HEAD. Fix: resolve_eval_worktree now applies the starting_patch first to reproduce the agent's base state before applying the candidate.

2. task_file.py (agent worktree): starting_patch often includes preprocessing artifacts (baseline_metrics.json, profile.json) that already exist in the fresh worktree, causing `git apply` to fail with "already exists in working directory". Fix: create_worktree_with_patch now falls back to copying kernel files directly from the patch's source worktree when git apply fails, instead of silently continuing with the unpatched base.

Closes #110

### DIFF
--- a/src/minisweagent/run/task_file.py
+++ b/src/minisweagent/run/task_file.py
@@ -10,6 +10,7 @@ CLI tools can create isolated work directories.
 
 from __future__ import annotations
 
+import logging
 import os
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


sdubagun-amd (changes requested):
Can you resolve the CI issues?

src/minisweagent/run/task_file.py
```diff
@@ -355,23 +367,86 @@ def create_worktree_with_patch(
...
+
+    try:
+        result = subprocess.run(
+            ["git", "worktree", "list"],
+            cwd=repo_path,
+            check=True,
+            capture_output=True,
+            text=True,
+            env=git_env,
+        )
```
sdubagun-amd: Can you add appropriate logger info messages to indicate which path the agent is taking?

src/minisweagent/run/task_file.py
```diff
@@ -355,23 +367,86 @@ def create_worktree_with_patch(
...
+            try:
+                subprocess.run(
+                    ["git", "worktree", "remove", str(worktree_path), "--force"],
+                    cwd=repo_path,
+                    check=True,
+                    capture_output=True,
+                    text=True,
+                    env=git_env,
+                )
+            except subprocess.CalledProcessError:
```
sdubagun-amd: Can we de-duplicate the two except statement codes?

sdubagun-amd (changes requested):

sdubagun-amd (approved):

Umangatamd (approved):

PR #134, opened 2026-04-11
fix: regenerate missing test script in worktree agents

iraj465:
When parallel agents run in worktrees, replace_paths() rewrites the _geak_test_cmd_*.sh script path from the original logs directory to the worktree path.  However, the script was only ever created in the original logs dir by _commandment_test_command(), so the rewritten path points to a non-existent file.

This causes save_and_test to fail with "script not found", making best_results.json report best_patch_speedup=1.0 for every task.  The orchestrator then cannot properly rank tasks within a round, potentially selecting a worse candidate for FULL_BENCHMARK evaluation.

Fix: add _ensure_test_script_exists() to save_and_test._run_test(). When the referenced _geak_test_cmd_*.sh doesn't exist, it regenerates the script from the COMMANDMENT sections (SETUP + CORRECTNESS + BENCHMARK) in the agent's working directory -- the same logic used by _commandment_test_command() in dispatch.py.


yueliu14 (commented):
I suggest copy *geak_test*.sh to orignal repo. We can also remove the hardcode **replace_path** of test command.

PR #136, opened 2026-04-12
feat(memory): enable in-session working memory for homogeneous mode

### LINKED ISSUE #130: Enable memory module in homogeneous agent
### Describe the issue

Currently, the memory module works only for heterogeneous agent. We want to enable it in homogeneous agent.

Umangatamd:
Closes #130

## Summary

In heterogeneous (Triton) mode the orchestrator explicitly creates a `WorkingMemory` instance and injects it into its LLM loop. In homogeneous (HIP) mode no such setup existed — each `StrategyInteractiveAgent` ran completely blind with no session memory.

## What changed

**3 files changed, `homogeneous_agent.py` untouched.**

### `mini.py`
Injects three `_wm_*` keys into `agent_config` before calling `run_homogeneous_agent()`:
- `_wm_baseline_metrics` → path to `baseline_metrics.json` in the preprocess output dir
- `_wm_benchmark_baseline` → path to `benchmark_baseline.txt`
- `_wm_kernel_path` → kernel file path from `preprocess_ctx`

These flow through `homogeneous_agent.py` unchanged (it just forwards `agent_config`).

### `parallel_agent.py`
In `run_single_agent()`:
- Pops the `_wm_*` keys from `parallel_agent_config` before passing `**parallel_agent_config` to the agent constructor (they are not valid constructor kwargs)
- Initializes a `WorkingMemory` instance per agent — identical pattern to `parallel_helpers.py` used by the heterogeneous pool runner — including:
  - Kernel category derivation via `classify_kernel_category(kernel_path)` with regex fallback
  - `load_baseline_from_artifacts()` to seed baseline latency and bottleneck type
  - `profiler_diagnosis` setup from `top_kernels` in `baseline_metrics.json`
- Sets `agent._working_memory = _wm`, which `default.py:query()` picks up on every LLM step and `_handle_tool_result()` picks up on every `save_and_test` / editor tool call

### `working_memory.py`
The `[SUBMIT NOW]` trigger message said _"Submit and let next round try differently"_ — orchestrator-specific language that doesn't apply to homogeneous mode (no rounds). Changed to _"Save your best patch and submit now."_

## How memory flows in homogeneous mode

```
mini.py
  └─ agent_config["_wm_*"] = artifact paths
       └─ run_homogeneous_agent()  [no changes]
            └─ ParallelAgent.run_parallel()
                 └─ run_single_agent() x N  [NEW: pops _wm_* keys, inits WorkingMemory]
                      └─ StrategyInteractiveAgent.run()
                           ├─ default.py:query()           → injects memory into LLM context each step
                           └─ default.py:_handle_tool_result() → extracts latency/speedup/edit insights
```

## Test plan

- [x] Run on HIP kernels end-to-end and confirm `[Working Memory]` blocks appear in agent conversation logs — verified in the Apr 21 follow-up comment: live `geak` run on `AgentKernelArena/tasks/hip2hip/others/knn` produced **13** `[Working Memory Update]` blocks in `traj.json` (0 before the `ParallelAgentConfig` fix)
- [x] Verify `GEAK_MEMORY_DISABLE=1` still suppresses memory (existing env gate)
- [ ] ~~Confirm no regression on heterogeneous (Triton) mode~~ — out of scope for this PR (no orchestrator/Triton code paths were changed; the fix only adds dataclass fields and homogeneous-only wiring)
- [x] 86 existing unit tests pass (`test_homogeneous_agent.py`, `test_working_memory.py`) — now **112/112** including `test_hip_benchmark_memory` and `test_mini`


Umangatamd:
Pushed a follow-up fix in commit 0e886a26 to address the HIP/null-memory issue from the reviewer report. I want to document the full reasoning here because the visible symptom (null values in the working-memory JSONL) was only the surface-level failure, and a smaller patch would have fixed the nulls while still producing the wrong overall metric.

## Root cause

The homogeneous in-session memory wiring from this PR was basically correct: homogeneous agents were already getting a WorkingMemory instance, they were recording strategy attempts/results, and the strategy manager still worked in homogeneous mode.

The real failure was in how different parts of GEAK interpreted benchmark output.

The reviewer's HIP artifacts were in a raw per-shape format like:

Perf: 0.2242 ms (shape_4_forward_backward)

Before this fix, the relevant parsers were inconsistent:

- working_notebook.parse_speedup_report() only knew how to extract numeric benchmark evidence from Overall: Xx (Y ms -> Z ms) or GEAK_RESULT_LATENCY_MS=...
- WorkingMemory baseline loading also only looked for GEAK_RESULT_LATENCY_MS=...
- WorkingMemory.note_tool_result() had a fallback that effectively used the last shape-like latency it could find in the output
- preprocessing/postprocessing had their own latency parsing paths, but that logic was duplicated and did not fully agree with the memory path

That caused two separate problems in the reviewer sample:

1. The notebook path could not parse the raw HIP Perf benchmark text, so the event log stored null overall_speedup, baseline_ms, and candidate_ms.

2. The baseline used by memory could silently fall back to the profiler-derived latency (1.528335 ms in the reviewer JSONL) instead of the actual benchmark aggregate, because the raw HIP benchmark baseline file did not contain a GEAK marker.

There was also a third, more subtle problem: even if I had only taught memory to grab the last Perf line, that would still have been wrong as an overall metric. In the reviewer sample, the last backward shape improved by about 22.44x, but the correct aggregate across the full reported benchmark shapes is about 2.406448x.

So a minimal patch that only made the nulls go away would still have overstated the overall win by almost an order of magnitude.

## Why the change is a bit broader

The diff ended up a little larger because benchmark parsing logic was duplicated in multiple places, and those copies had already drifted.

If I had fixed only the memory notebook parser, we would still have had:

- one interpretation of HIP output in memory
- another interpretation in deterministic patch selection
- another interpretation in preprocessing-generated harness output

That would make this class of bug very likely to come back the next time a backend or harness format changes.

So instead of patching only one local regex, I centralized the benchmark parsing rules and then pointed the memory path and the deterministic evaluation path at the same source of truth.

## What changed

1. Added src/minisweagent/run/benchmark_output.py as the shared parser for benchmark latency, per-shape latencies, benchmark config identifiers, and per-shape speedups.

2. Added explicit support for raw HIP Perf lines and generic shape-labeled latency lines.

3. Defined the canonical overall latency for raw multi-shape outputs without a trustworthy explicit summary as the geometric mean of the parsed per-shape latencies.

4. Updated WorkingMemory baseline loading so it stores both the overall benchmark baseline latency and the per-shape baseline latencies instead of only looking for a GEAK marker.

5. Updated working_notebook.parse_speedup_report() so it can compute both overall and per-shape speedups from raw HIP Perf outputs using the stored baseline shape map.

6. Updated the generated C-like wrapper path in harness_utils.py so that when it needs to infer a benchmark latency from raw Perf output, it emits the geomean across parsed shapes rather than only matching a single launch-style line.

7. Updated the preprocess and postprocess benchmark parsing modules to re-export the shared benchmark parser helpers instead of carrying their own subtly different copies.

## Concrete validation on the reviewer sample

I replayed the reviewer-style HIP output through the new parser and memory path.

For the sample full benchmark baseline and improved candidate output, the parser now computes:

- baseline_ms = 0.190861
- candidate_ms = 0.079312
- overall_speedup = 2.406448x

And importantly, we still preserve the strong single-shape improvement:

- shape_4_forward_backward = 22.439786x

So after the fix, we keep both pieces of information correctly:

- the real overall benchmark result
- the big per-shape localized win

This is exactly the distinction that was getting lost before.

## Tests added and updated

Added tests/test_hip_benchmark_memory.py to cover:

- raw HIP Perf output parsing
- geomean latency computation from shape lines
- shape label extraction for HIP output
- notebook speedup parsing from raw HIP output
- WorkingMemory baseline loading and speedup tracking on HIP output

Updated tests/test_working_memory.py so shape-only outputs now use the geomean rather than the last-shape latency.

Ran:

pytest tests/test_hip_benchmark_memory.py tests/test_benchmark_parsing.py tests/test_working_memory.py

Result:

62 passed

## Net effect for this PR

After this follow-up commit, the homogeneous in-session memory path should now behave correctly for the reviewer’s HIP case:

- no null benchmark fields in the notebook/event log for raw HIP benchmark outputs
- no accidental fallback to profiler latency when a benchmark baseline exists but is only expressed as raw Perf lines
- no inflated overall speedup from using the final shape as the aggregate metric
- shared benchmark parsing behavior across preprocess, postprocess, and memory paths

If anyone reviewing this wants, I can also paste a smaller before/after example from the reviewer artifact bundle showing the exact transition from null and last-shape behavior to the corrected aggregate plus per-shape breakdown.

Umangatamd:
## Root cause + fix for the "agent runs blind in HIP mode" report

While exercising this branch end-to-end on HIP kernels (`knn`, `device_merge_sort`), we noticed that agent trajectories contained **zero** `[Working Memory Update]` blocks even though the whole purpose of this PR is to inject them into homogeneous-mode sub-agents. Digging in, the three `_wm_*` artefact keys that `run/mini.py` stuffs into `agent_config` never make it to `run_single_agent.pop(...)`.

### Why — two back-to-back dataclass losses

`agent_config` has to traverse `ParallelAgentConfig` twice:

1. `homogeneous_agent.py:138`
    ```python
    agent = ParallelAgent(model, env, **agent_config)
    ```
2. `parallel_agent.py:97-101` (inside `ParallelAgent.run`)
    ```python
    agent_config={k: v for k, v in self.config.__dict__.items() if k not in (...)}
    ```

`ParallelAgentConfig` is a `@dataclass` with **no** `_wm_*` fields, so step 1 raises
```
TypeError: ParallelAgentConfig.__init__() got an unexpected keyword argument '_wm_baseline_metrics'
```
and step 2 would silently drop the keys even if step 1 were survived. Result: homogeneous sub-agents crash at construction, `_working_memory` is never attached, and no `[Working Memory Update]` ever reaches the LLM.

### Fix — declare the fields on the dataclass

Latest commit adds:
```python
# Working-memory artefact paths (set by run/mini.py for homogeneous mode so that
# run_single_agent can initialise a per-slot WorkingMemory).
_wm_baseline_metrics: str | None = None
_wm_benchmark_baseline: str | None = None
_wm_kernel_path: str | None = None
```
on `ParallelAgentConfig`. That's it — now both traversals carry the values, and `run_single_agent`'s existing pop works.

Also included: a small `Dockerfile` cleanup removing the stale top-level `COPY knowledge-base/ knowledge-base/` line (the directory was moved under `mcp_tools/rag-mcp/knowledge-base/` in d6bbd9c7 and is already pulled in by the `mcp_tools/` COPY). The branch couldn't `docker build` without this; same issue exists on `main`.

### Verification

End-to-end run of `geak` on `AgentKernelArena/tasks/hip2hip/others/knn` (GPU 0, 1 parallel agent):

| check | before | after |
|---|---|---|
| homogeneous agent constructs | ❌ `TypeError` | ✅ |
| `_working_memory/events/homo-agent-0.jsonl` present | ❌ | ✅ baseline + step events |
| baseline latency / bottleneck / per-shape latencies loaded | ❌ | ✅ `0.2451 ms`, `compute`, 15 shapes |
| `[Working Memory Update]` blocks in `traj.json` | **0** | **13** (in ~3.5 min of real run) |
| profiler-dominant-kernel diagnosis fires | ❌ | ✅ `[PROFILER] Target kernel (knn_kernel) dominates at 98%` |
| existing unit tests (`test_homogeneous_agent`, `test_working_memory`, `test_hip_benchmark_memory`, `test_mini`) | pass | pass (112/112) |

Also rebased the branch onto latest `main` while I was in here (was 11 behind — clean rebase, no conflicts). Feel free to revert the rebase piece if you prefer to merge the fix on its own.

PR #146, opened 2026-04-13
fix(preprocess): use full Metrix profile for baseline runs

### LINKED ISSUE #163: Weakly wired profiler information
### Describe the bug

The GEAK optimizer uses a weakly consolidated profiler information:
1. It currently aggregates metrics of all the kernels triggered during evaluation, including runtime/driver kernels, framework dispatch helpers, autograd-related, etc: the useful kernels targeted by the optimization may sometimes be a small fraction of them.
2. It currently gives information about some, but not all the fields computed during profiling.

Umangatamd:
## Summary
- switch preprocess baseline profiling from Metrix quick mode to the full memory profile so baseline artifacts include the full hardware counter set
- update both preprocess entry paths plus the shared baseline profiling helper to keep Metrix behavior consistent
- leave benchmark-duration override behavior unchanged so downstream scoring still uses the intended canonical wall-clock metric

## Test plan
- [x] `python3 -m py_compile src/minisweagent/run/preprocess/preprocessor.py src/minisweagent/run/preprocess/harness_utils.py src/minisweagent/run/pipeline_helpers.py`
- [x] `pytest tests/run/test_pipeline_helpers.py -q`
- [x] `pytest tests/run/test_preprocess_unit.py -q`

### DIFF
--- a/src/minisweagent/run/pipeline_helpers.py
+++ b/src/minisweagent/run/pipeline_helpers.py
@@ -903,6 +903,6 @@ def run_baseline_profile(test_command: str, gpu_id: int = 0) -> dict:
         command=profile_cmd,
         backend="metrix",
         num_replays=3,
-        quick=True,
+        quick=False,
         gpu_devices=str(gpu_id),
     )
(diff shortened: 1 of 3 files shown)


sdubagun-amd (approved):
Looks good as of now for main. There is one potential source where the method `_run_patch_profile()` still does quick profiling, in the file `tools/save_and_test.py`. This is only for doing profiling per each patch, which is not set by default and not wired properly. This is just a note, so we won't lose track of it.

iraj465 (approved):

Issue #151, opened 2026-04-15
Tools Set Unifiy


yueliu14:
### Describe the issue

Currently, the Tools set loadings are not unified. We should unified Tools Set for agents.

PR #152, opened 2026-04-15
Fix/unified toolruntime

chao-xu-spec:
#151 
1. Unify the `ToolRuntime` contract: `ToolRuntime` owns assembling native and MCP tools and keeps _tool_table aligned with `dispatch()`. Expose the model-facing OpenAI-style definitions via `get_tools_list()`, remove tools by name via `disable_tools()`, and provide `fetch_tools_list()` to materialize the same `tools_list` without retaining a long-lived `ToolRuntime` instance (for one-off schema retrieval).
2. Faster MCP tool discovery without listing-time stdio (default layout): When `server_config` matches the built-in in-repo `mcp_tools/<server>/ defaults`, tool discovery can use an in-process import plus `mcp.list_tools()` instead of spawning a subprocess just to enumerate tools. Actual tool execution still uses the stdio `MCPClient` path for isolation and parity with real calls.
3. Remove`metrix-mcp`: the `metrix` backend  of `profiler-mcp` provides the same functionality.
4. Remove `ToolRuntime` ownership from `StrategyAgent`: `StrategyAgent` no longer constructs a `ToolRuntime`. `DefaultAgent` (via the inheritance chain) owns runtime construction, applies disabled_tools, and syncs `toolruntime.get_tools_list()` to the model; `StrategyAgent` keeps strategy-list concerns and tool_profile-related wiring only.
5. Decouple models from tool policy: `amd_llm`, `litellm_model` models, etc. only accept `set_tools(tools_list)` prepared upstream. They do not instantiate `ToolRuntime`, interpret `tool_profile`, or apply `disable_tools`; the agent + `ToolRuntime` are the single source of truth for which tools exist before injection into the model.
6. Collapse end-user profiling into `profiler-mcp`: Remove the standalone profiling tool from the agent-visible surface. Profiling is exposed as `profile_kernel` on `profiler-mcp` (covering Metrix and rocprof-compute backends). Rocprof-specific helpers may remain as libraries for the backend implementation but are not a separate LLM-callable tool name.
7. Adapt heterogeneous and homogeneous flows to `ToolRuntime`: Update orchestration, task generation, schemas, prompts, and parallel/homogeneous paths so they rely on the unified ToolRuntime table and get_tools_list() output, avoiding divergent tool sets between orchestration and DefaultAgent.
8. `mini.py` + `geak.yaml` tool toggles: After merging YAML, mini.py maps every key under tools: whose value is boolean false into agent.disabled_tools (generic loop, not a hard-coded pair of names). The sample `geak.yaml` uses profile_kernel: false to match the real registered tool name.
9. `profiler-mcp`: require backend on `profile_kernel`: The backend argument is required and restricted to `metrix` or `rocprof-compute` (reflected in MCP schemas as required + enum), forcing the agent to choose the backend explicitly on each call and removing implicit defaults that could hide backend choice.


yueliu14 (approved):

sdubagun-amd (approved):
LGTM

PR #158, opened 2026-04-16
fix(preprocess): harden UTA, shape fixer, and baseline reruns

### LINKED ISSUE #154: Shape Fixer Agent Infinite Loop — No Termination Condition
### Bug Summary
The shape fixer agent (`shape_fixer_agent.py`) enters an infinite loop when it prints `SHAPES_FIXED` because the mini-swe-agent framework has no mechanism to terminate on that keyword. The agent keeps running indefinitely, consuming LLM API tokens with zero useful output.
**Observed**: rotary_embedding kernel — shape fixer ran **1,565 steps** (16,153 lines of log, 90+ minutes) printing `SHAPES_FIXED` repeatedly before being manually killed.

> Got this error while running GEAK (3d6311e877fa121a01e2ab528c5e51e23da42261) in Hyperloom setting.
### The Bug (in `shape_fixer_agent.py`, lines `115-121`)
```
exit_status, result = agent.run(task)       # Line 115
if "SHAPES_VERIFIED" in (result or ""):     # Line 117
    return True
if "SHAPES_FIXED" in (result or ""):        # Line 119
    return True
return False                                # Line 121
```
This function calls `agent.run(task)` which starts the LLM agent. The agent's system prompt says:

> "Print SHAPES_FIXED. Stop."

But `agent.run()` is mini-swe-agent's main loop which **doesn't stop when the LLM prints SHAPES_FIXED**. The agent only stops when:

1. It hits step_limit (set to 0.0 = unlimited)
2. It hits cost_limit (set to 0.0 = unlimited)
3. It outputs a `MINI_SWE_AGENT_FINAL_OUTPUT` token
4. The model produces no more actions

The system prompt tells the LLM to "Print SHAPES_FIXED. Stop." but the LLM doesn't actually know how to stop the agent loop. It just prints `SHAPES_FIXED` as text, and the agent framework keeps asking it for the next action. The LLM, having nothing left to do, prints `SHAPES_FIXED` again. And again. 1,515 times.

### Proof from the log
Step 52 (line 979) — the agent correctly identifies the problem and fixes the harness:
```
The file is correct. Let me do a final summary verification:
...
SHAPES_FIXED
```
Step 53 (line 1002) — agent has nothing to do, prints the keyword again:
```
SHAPES_FIXED
```
Step 54 (line 1006) — same thing, with a longer explanation this time:
```
SHAPES_FIXED. The harness has been corrected. Here is the summary of what was wrong...
```
Steps 55-1565 — pure loop, `SHAPES_FIXED` repeated 1,515 times with empty user responses (`<returncode>0</returncode><output></output>`).

### Why `topkGatingSoftmax` didn't loop
The topkGatingSoftmax shape fixer log has 291 steps (vs 1,565 for rotary). Looking at the step count, it likely hit some implicit limit or the shapes matched earlier (requiring SHAPES_VERIFIED which may have triggered a slightly different code path). But the fundamental issue is the same — there's no proper termination signal.

### Why `step_limit: 0.0` means unlimited
From `shape_fixer_agent.py` line 91:
```
"step_limit": 0.0,
"cost_limit": 0.0,
```
In mini-swe-agent, `0.0` means _no limit_. The agent will run until the model stops producing actions. But when the model prints `SHAPES_FIXED` (plain text, not a tool call), the framework interprets this as "the model produced output" and prompts it again. The model never produces "no action" because it keeps responding to the prompt.

#### The fix would be
The `run_shape_fixer` function should either:

1. Set a finite step_limit (e.g., 50 steps)
2. Have the agent use `MINI_SWE_AGENT_FINAL_OUTPUT` instead of bare `SHAPES_FIXED`
3. Check the result after each step and break if SHAPES_FIXED or SHAPES_VERIFIED appears
This is a GEAK-side bug in `shape_fixer_agent.py` — the agent has no termination condition and relies on the LLM "stopping" itself, which it can't do within the mini-swe-agent framework.

### Impact
1. 90+ minutes wasted with zero useful output (harness was already fixed at step 52)
2. 1,515 unnecessary LLM API calls to claude-opus-4-6 (at ~$0.27/step = ~$400+ burned)
3. The rotary_embedding kernel never reached profiling, baseline, or optimization — all downstream steps were blocked
4. On a subsequent run (topkGatingSoftmax), the shape fixer loop was compounded by a find / full filesystem scan, causing the container to hang for 3+ hours

### LINKED ISSUE #156: UTA generates `find /` full filesystem scan when INSTRUCTIONS.md path is unresolvable in non-GEAK WORKDIR containers
## Bug Summary

The UTA (Unit Test Agent) generates a `find /` command that scans the entire container filesystem when it cannot locate `INSTRUCTIONS.md` at the path specified in the task prompt. On containers with large NFS-mounted volumes, this hangs for hours in uninterruptible I/O (D+ state).

**Observed**: topkGatingSoftmax kernel — UTA step 13 ran `find / -path "*/minisweagent/run/preprocess/INSTRUCTIONS.md"` which hung for 3+ hours before being manually killed. Process was in D+ state (uninterruptible I/O on NFS).

> Got this error while running GEAK (3d6311e877fa121a01e2ab528c5e51e23da42261) in Hyperloom setting.
> Here my `WORKDIR=/home/user` instead of `/path/to/GEAK/`.

## Root Cause

There are two compounding issues:

### 1. The task prompt references a path relative to the GEAK repo, not the `WORKDIR` location as modified in `GEAK/Dockerfile`.

In `/src/minisweagent/run/preprocess/unit_test_agent.py` (line 215), the task prompt tells the UTA:

```
IMPORTANT: Read src/minisweagent/run/preprocess/INSTRUCTIONS.md in the repository for test harness requirements and COMMANDMENT format rules before creating the harness.
```

The "repository" here is the **kernel's repo** (e.g., `/sgl-workspace/sglang/sgl-kernel/csrc/moe`), NOT the GEAK repo. But `src/minisweagent/run/preprocess/INSTRUCTIONS.md` only exists inside the GEAK repo (`/workspace` in GEAK's Dockerfile, or `/home/user/PROJECTS_NEW/GEAK` in the Hyperloom Dockerfile).

The UTA agent receives `Repository: /sgl-workspace/sglang/sgl-kernel/csrc/moe` and tries:

1. `cat /sgl-workspace/sglang/sgl-kernel/csrc/moe/src/minisweagent/run/preprocess/INSTRUCTIONS.md` — **NOT FOUND** (correct, it's not there)
2. `find /sgl-workspace/sglang ... -path "*/INSTRUCTIONS.md"` — **NOT FOUND** (correct, still not there)
3. `find / -path "*/minisweagent/run/preprocess/INSTRUCTIONS.md"` — **HANGS FOREVER**

The LLM escalates from scoped search to full filesystem scan as a fallback. This is an LLM reasoning problem triggered by asking it to find a file that doesn't exist in the repo it was given.

### 2. WORKDIR mismatch between GEAK Dockerfile and Hyperloom Dockerfile

**GEAK's Dockerfile** (`GEAK/Dockerfile`):
```dockerfile
WORKDIR /workspace
COPY . .
RUN pip install -e .
```

GEAK is installed at `/workspace`. The path `src/minisweagent/run/preprocess/INSTRUCTIONS.md` resolves to `/workspace/src/minisweagent/run/preprocess/INSTRUCTIONS.md`.

Hyperloom's Dockerfile (MODIFIED) (`GEAK/Dockerfile`):

```
COPY GEAK /home/user/PROJECTS_NEW/GEAK
WORKDIR /home/user/PROJECTS_NEW/GEAK
RUN pip install -e '.[dev,langchain]'
WORKDIR /home/user               # <-- final WORKDIR
```
GEAK is installed at `/home/user/PROJECTS_NEW/GEAK`. The `INSTRUCTIONS.md` is at `/home/user/PROJECTS_NEW/GEAK/src/minisweagent/run/preprocess/INSTRUCTIONS.md`.

Hyperloom's `scripts/run-docker.sh` (line 233):
```
-w /home/user    # container working directory
```

When `docker exec` runs GEAK, CWD is `/home/user` (a mounted volume containing the entire home directory). The UTA's `find /` then walks:

- `/home/user` — the entire mounted home directory (potentially terabytes on NFS)
- `/cephfs` — mounted shared storage
- `/shared-nfs` — mounted NFS
- `/shared-aig` — mounted shared storage
- `/mnt` — mounted storage
All specified in `scripts/run-docker.sh` lines 227-232:
```
-v /cephfs:/cephfs \
-v "${PARENT_DIR}:${PARENT_DIR}" \
-v /mnt:/mnt \
-v /shared-nfs:/shared-nfs \
-v /shared-aig:/shared-aig \
```
The chain of events
```
1. unit_test_agent.py builds task prompt:
   "Read src/minisweagent/run/preprocess/INSTRUCTIONS.md in the repository"
   
2. Repository = /sgl-workspace/sglang/sgl-kernel/csrc/moe (the kernel repo)
3. INSTRUCTIONS.md does NOT exist in the kernel repo
   (it's in /home/user/PROJECTS_NEW/GEAK/src/...)
4. UTA (LLM) tries: cat <repo>/src/minisweagent/... → NOT FOUND
5. UTA (LLM) tries: find <repo> -path "*/INSTRUCTIONS.md" → NOT FOUND
6. UTA (LLM) escalates: find / -path "*/minisweagent/run/preprocess/INSTRUCTIONS.md"
   → scans /home/user (NFS mount, terabytes)
   → scans /cephfs, /shared-nfs, /shared-aig, /mnt
   → process enters D+ state (uninterruptible I/O)
   → hangs for 3+ hours
```
Proof from Logs
From the terminal session (topkGatingSoftmax run):

`root  8647  0.0  0.0  0  0  ?  D+  06:05  0:00  find / -path */minisweagent/run/preprocess/INSTRUCTIONS.md`
- PID 8647 had been running for 2 hours 46 minutes at time of inspection
- Process state: `D+` (uninterruptible I/O — stuck reading from slow/huge filesystem)
- The `find /` was generated at UTA step 13

The UTA log confirms the escalation:
```
Step 12: cat /sgl-workspace/sglang/sgl-kernel/csrc/moe/src/minisweagent/run/preprocess/INSTRUCTIONS.md
→ NOT FOUND
Step 13: find /sgl-workspace/sglang ... -path "*/INSTRUCTIONS.md" 2>/dev/null; find / -path "*/minisweagent/run/preprocess/INSTRUCTIONS.md" 2>/dev/null | head -5
→ HUNG (the second find / never returns)
```

### Impact
- 3+ hours wasted per occurrence — the `find /` never terminates on containers with large NFS mounts
- GEAK never reaches harness creation, profiling, or optimization
- The process is in D+ (uninterruptible) state — cannot even be killed with `kill -9` in some cases; requires killing the parent process or the container

#### Resolution:
Resolve `INSTRUCIONS.md` path at build time
```
import minisweagent
geak_root = Path(minisweagent.__file__).parent.parent  # or use importlib.resources
instructions_path = geak_root / "run" / "preprocess" / "INSTRUCTIONS.md"

task = (
    f"Create a fixed test harness for kernel: {kernel_name}\n"
    f"Repository: {repo}\n\n"
    f"IMPORTANT: Read {instructions_path} for test harness requirements\n"
    f"and COMMANDMENT format rules before creating the harness."
)
```

### LINKED ISSUE #157: `--iterations 30` flag crashes every UTA-generated harness baseline re-run (silent degradation)
### Bug Summary

The GEAK preprocessor (`preprocessor.py`) unconditionally appends `--iterations 30` to the harness CLI command for baseline re-runs, but the UTA-generated harness's argparse doesn't accept `--iterations` as a flag. This causes the `--benchmark` mode to crash with exit code 2 on every baseline re-run, silently degrading baseline quality and downstream evaluation.

**Observed**: All 3 kernels in the 2026-04-14 run (`ck_moe_stage1`, `ck_moe_stage2`, `paged_attention`) and `topkGatingSoftmax` in the 2026-04-16 run — `--benchmark FAILED (rc=2)` during baseline re-run with `unrecognized arguments: --iterations 30`.

> Got this error while running GEAK (3d6311e877fa121a01e2ab528c5e51e23da42261) in Hyperloom setting.

### The Bug (call chain across 3 files)
```
preprocessor.py:949    →  extra = "--iterations 30"
    ↓
harness_utils.py:1142  →  sets env: GEAK_BENCHMARK_EXTRA_ARGS="--iterations 30"
    ↓
run_harness.py:85-87   →  appends "--iterations 30" to the harness command
    ↓
python test_harness.py --benchmark --iterations 30
    ↓
💥 harness's argparse rejects --iterations → exit code 2
```

After the shape fixer completes and all four harness modes pass (`--correctness`, `--profile`, `--benchmark`, `--full-benchmark`), the preprocessor re-runs the harness with `--iterations 30` to collect stable baseline measurements with more repetitions. But the UTA-generated harness defines its argparse as:

```python
parser.add_argument("--correctness", ...)
parser.add_argument("--benchmark", ...)
parser.add_argument("--full-benchmark", ...)
parser.add_argument("--profile", ...)
# No --iterations argument
```
The `--iterations` flag is a GEAK preprocessor convention that the UTA agent doesn't know about. The harness uses a hardcoded iteration count (or reads `GEAK_BENCHMARK_ITERATIONS` from the environment), but never exposes it as a CLI argument.

### Proof from the log
The four harness modes pass cleanly:

```
run_harness: --correctness passed (1.3s)
run_harness: --profile passed (1.2s)
run_harness: --benchmark passed (1.2s)
run_harness: --full-benchmark passed (1.4s)
Re-validation after shape fix: ALL MODES PASSED
```
Then the baseline re-run with `--iterations 30` is triggered:
```
Re-running all modes with --iterations 30 for baselines...
run_harness: --correctness passed (1.3s)
run_harness: --profile passed (1.2s)
run_harness: --benchmark FAILED (rc=2, 1.0s)
```
With the error:
```
usage: test_moe_topk_softmax_kernels_harness.py [-h]
    (--correctness | --profile | --benchmark | --full-benchmark)
test_moe_topk_softmax_kernels_harness.py: error: unrecognized arguments: --iterations 30
```
Note that `--correctness` and `--profile` pass because `run_harness.py` only appends `--iterations 30` to the `--benchmark` and `--full-benchmark` modes (lines 85-87), not to correctness/profile.

### Why it's silent — GEAK continues despite the failure
From the log, after the crash:

```
WARNING: baseline re-run had failures: ['--benchmark mode failed (exit code 2):...']
```
The preprocessor logs a WARNING but does not abort. It proceeds to Step 5 (profiling), Step 6 (baseline metrics from the earlier successful run), and Step 7 (commandment generation). The pipeline continues because:

1. The initial harness run (without `--iterations 30`) already passed — those results are cached
2. The `--iterations 30` re-run is an enhancement for higher-confidence baselines, not a hard requirement
3. `preprocessor.py` treats this as a non-fatal warning

### What it breaks
|What|How|
|---|---|
|benchmark_baseline.txt|Never saved — the 30-iteration high-confidence baseline doesn't exist|
|Absolute baseline comparison in evaluation|Warns "No full benchmark baseline found" — evaluation can't compare patches against a stable absolute baseline|
|Strategy manager confidence|Marks some strategies as "partial" instead of "successful" due to missing baseline reference|
|Baseline variance|The baseline used for patch ranking comes from a single run (default iterations), not the intended 30-iteration averaged baseline. Patch speedup measurements have higher variance than intended.|

### What it doesn't break
|What|Why|
|---|---|
|Relative patch-vs-patch speedup|All patches are evaluated with the same harness and same iteration count — relative ranking is consistent|
|Correctness checks|`--correctness` mode is unaffected (no `--iterations` appended)|
|The run producing results|GEAK continues past the warning and produces `best_results.json` with valid patch speedups|
|Profiling (Metrix)|Uses `--profile` mode, which is also unaffected|

### Why this happens for every UTA-generated harness
The UTA agent generates the harness argparse from scratch each time. It knows about the four modes (`--correctness`, `--benchmark`, `--full-benchmark`, `--profile`) because those are specified in the GEAK harness contract (`INSTRUCTIONS.md`). But `--iterations` is not part of the harness contract — it's an internal preprocessor convention added via `GEAK_BENCHMARK_EXTRA_ARGS`. The UTA has no way to know it needs to accept this flag.

The mismatch is:

- Preprocessor assumes: the harness CLI accepts `--iterations N`
- UTA generates: a harness that only accepts the four documented modes
- Harness contract (`INSTRUCTIONS.md`): doesn't mention `--iterations`

#### The possible fix (not implemented)
Pass the iteration count only via the `GEAK_BENCHMARK_ITERATIONS` environment variable (which harnesses already read internally), and stop appending `--iterations` to the CLI command. Specifically in `run_harness.py` lines 85-87, replace the CLI argument injection with:
```python
env["GEAK_BENCHMARK_ITERATIONS"] = str(iterations)
# Don't append --iterations to cmd
```
This way the harness reads the iteration count from the environment (already supported) and argparse never sees an unknown flag.

### Impact
1. Every UTA-generated harness triggers this warning — 100% reproduction rate across all kernels tested (ck_moe_stage1, ck_moe_stage2, paged_attention, topkGatingSoftmax)
2. Baseline quality is silently degraded — single-run baselines instead of 30-iteration averaged baselines
3. No LLM cost or time waste (unlike the shape fixer loop) — the crash is instant (rc=2, ~1s), and the pipeline continues
4. Patch speedup measurements are noisier than necessary, which could cause GEAK to select a suboptimal patch when speedups are close (e.g., the ck_moe_stage1 patches ranged 1.003x–1.011x — a noisy baseline could flip the ranking)

Umangatamd:
## Summary
- Close #154 by making the shape fixer stop on `SHAPES_VERIFIED`/`SHAPES_FIXED`, adding a wall-clock timeout, bounding the mini shape-fixer budget, and restoring the original harness if a fix attempt fails or produces an invalid harness.
- Close #156 by telling the Unit Test Agent the exact GEAK-relative path for `src/minisweagent/run/preprocess/INSTRUCTIONS.md` and explicitly forbidding broad filesystem scans such as `find /` and `find .`.
- Close #157 by normalizing `--iterations N` into `GEAK_BENCHMARK_ITERATIONS` so baseline reruns no longer depend on harnesses exposing an `--iterations` CLI flag, while preserving any non-iteration benchmark extra args.
- Add regression coverage for the new UTA prompt guidance, shape-fixer termination, harness restoration, and benchmark extra-arg normalization.

## Validation
- Ran `pytest tests/run/test_preprocess_unit.py`.
- Ran a preprocess-focused smoke test in the `geak-agent-upandey` container against the `topkGatingSoftmax` HIP kernel flow.
- Ran a full `geak -t ... --yolo --exit-immediately` smoke test in the same container for `topkGatingSoftmax`, monitored it through preprocessing, and stopped it immediately after preprocessing completed.
- Result: the UTA no longer performed broad `find` scans, the shape fixer no longer hung the pipeline and failed open when needed, the preprocessing flow generated the expected artifacts (`profile.json`, `baseline_metrics.json`, `COMMANDMENT.md`), and the baseline rerun no longer crashed on an unsupported `--iterations 30` CLI flag.

## Issues
Closes #154.
Closes #156.
Closes #157.

### DIFF
--- a/src/minisweagent/run/pipeline_helpers.py
+++ b/src/minisweagent/run/pipeline_helpers.py
@@ -388,10 +388,10 @@ def execute_harness_validation(
     Parameters
     ----------
     benchmark_extra_args:
-        Extra CLI args appended to benchmark/full-benchmark invocations
-        (e.g. ``"--iterations 50"``).  Passed via the
-        ``GEAK_BENCHMARK_EXTRA_ARGS`` env var so both direct invocations
-        and COMMANDMENT-based scripts use the same settings.
+        Extra benchmark tuning args. ``--iterations N`` is normalized into
+        ``GEAK_BENCHMARK_ITERATIONS`` so harnesses do not need to expose it
+        as a CLI flag. Any remaining args are passed via
+        ``GEAK_BENCHMARK_EXTRA_ARGS``.
 
     Returns
     -------
@@ -410,15 +410,19 @@ def execute_harness_validation(
     if not benchmark_extra_args:
         env_overrides["GEAK_BENCHMARK_ITERATIONS"] = "5"
     else:
-        env_overrides["GEAK_BENCHMARK_EXTRA_ARGS"] = benchmark_extra_args
-        # Extract --iterations N from extra_args and also set the env var
-        # so harnesses that read GEAK_BENCHMARK_ITERATIONS (instead of
-        # parsing --iterations from argv) get the right count.
         import re as _re
 
-        _iter_match = _re.search(r"--iterations\s+(\d+)", benchmark_extra_args)
+        remaining_extra_args = benchmark_extra_args.strip()
+        _iter_match = _re.search(r"(?:^|\s)--iterations\s+(\d+)(?=\s|$)", remaining_extra_args)
         if _iter_match:
             env_overrides["GEAK_BENCHMARK_ITERATIONS"] = _iter_match.group(1)
+            remaining_extra_args = _re.sub(
+                r"(?:^|\s)--iterations\s+\d+(?=\s|$)",
+                " ",
+                remaining_extra_args,
+            ).strip()
+        if remaining_extra_args:
+            env_overrides["GEAK_BENCHMARK_EXTRA_ARGS"] = remaining_extra_args
 
     results = run_harness(
         harness_path,
--- a/src/minisweagent/run/preprocess/INSTRUCTIONS.md
+++ b/src/minisweagent/run/preprocess/INSTRUCTIONS.md
@@ -199,6 +199,17 @@ script that imports the kernel, creates test inputs, and provides
    x = torch.randn(S, B, H, D, dtype=torch.float16, device='cpu').to('cuda')
    ```
 
+9. **Preserve the source benchmark/test's full execution contract, not just
+   shapes.**
+   - Keep per-tensor semantics independent: dtype, device, layout,
+     contiguity, index dtypes, auxiliary buffers/caches/scales, and any
+     helper-side preprocessing.
+   - Do NOT normalize every tensor to the main activation dtype just because
+     the benchmark uses that dtype for `query`, `key`, or other primary
+     activations.
+   - If a harness mode fails, understand why it fails before editing. Prefer
+     the smallest source-faithful fix over rewriting working harness logic.
+
 **Language-specific test harness notes:**
 
 When the kernel is NOT a Python/Triton kernel, the test harness approach varies:
--- a/src/minisweagent/run/preprocess/config/mini_shape_fixer.yaml
+++ b/src/minisweagent/run/preprocess/config/mini_shape_fixer.yaml
@@ -1,6 +1,8 @@
 agent:
   system_template: |
-    You are a meta-checking agent. Read two files, compare shapes, done.
+    You are a harness repair agent. Start by comparing the authoritative
+    source file against the generated harness, then apply the smallest
+    source-faithful fix needed to make the sampled modes work.
 
     Step 1: Read the SHAPE SOURCE FILE. What configs does it benchmark?
       (Look for the config variables, loops, products, helper functions, and
@@ -20,11 +22,14 @@ agent:
       Then determine the exact sampled cases that the harness will use for
       `--benchmark`, `--correctness`, and `--profile`.
 
-    Step 3: Compare EXACT sampled semantics, not just the config universe.
+    Step 3: Compare full sampled semantics, not just the config universe.
       - SAME VALUES / SAME COUNT is NOT sufficient if a different order causes
         `_pick()` (or equivalent logic) to choose different sampled cases.
-      - If the source file already defines a case list/helper, that ordering is
-        authoritative.
+      - Preserve the source file's helper semantics and per-tensor contracts:
+        dtype, device, layout, contiguity, auxiliary buffers/caches/scales,
+        index dtypes, and helper-side preprocessing.
+      - Do NOT normalize all tensors to a single dtype just because the main
+        activations use that dtype.
       - If a task-runner exists, preserve it only insofar as it faithfully
         wraps the same repo-native benchmark/test semantics. Do NOT approve a
         wrapper that invents new commands, unsupported flags, or a different
@@ -32,12 +37,21 @@ agent:
       - If the source file uses nested loops, preserve that loop nesting order.
       - Tuple vs list syntax is fine only when the resulting sampled cases are
         exactly the same.
-      - YES: exact sampled cases for `--benchmark`, `--correctness`, and
-        `--profile` all match the source file. Print SHAPES_VERIFIED. Stop.
-      - NO: fix the harness so it preserves the source ordering and yields the
-        exact same sampled cases. Print SHAPES_FIXED. Stop.
 
-    That is all. Do not run anything. Do not explore. Just read and compare.
+    Step 4: Validate and repair minimally.
+      - If validation commands are provided, you MAY run them from the repo
+        root to understand why a mode fails.
+      - Understand failures from stderr before editing.
+      - Re-run only the relevant failing mode(s) after each edit.
+      - Prefer the smallest source-faithful fix over rewriting working harness
+        logic.
+      - If no edit is needed and the relevant validation commands pass, print
+        SHAPES_VERIFIED. Stop.
+      - If you edited the harness and the relevant validation commands pass,
+        print SHAPES_FIXED. Stop.
+
+    Stay focused: read the source file, the harness, and directly imported
+    helper files when needed. Do not broad-explore the repo.
   step_limit: 80
   cost_limit: 5.0
   instance_template: |
--- a/src/minisweagent/run/preprocess/config/mini_shape_fixer.yaml
+++ b/src/minisweagent/run/preprocess/config/mini_shape_fixer.yaml
@@ -38,8 +38,8 @@ agent:
         exact same sampled cases. Print SHAPES_FIXED. Stop.
 
     That is all. Do not run anything. Do not explore. Just read and compare.
-  step_limit: 0.0
-  cost_limit: 0.0
+  step_limit: 80
+  cost_limit: 5.0
   instance_template: |
     Your task is: {{task}}
   action_observation_template: |
--- a/src/minisweagent/run/preprocess/config/mini_unit_test_agent.yaml
+++ b/src/minisweagent/run/preprocess/config/mini_unit_test_agent.yaml
@@ -44,11 +44,13 @@ agent:
        This resolves all dependency issues upfront. Do NOT manually search for packages.
     3) Your task context contains **pre-scanned discovery results** from an automated scan.
        Review the Kernel Analysis, discovered test files, and language-specific guidance provided.
-    4) **If `src/minisweagent/run/preprocess/INSTRUCTIONS.md` exists** in the repository, read sections 1a and 1b for:
+    4) GEAK's harness instructions live at the repo-relative path
+       `src/minisweagent/run/preprocess/INSTRUCTIONS.md` inside the GEAK repo.
+       This file may not exist inside the target kernel repository.
+       If you can read that exact relative path directly, use sections 1a and 1b for:
        - Test harness requirements (the `--correctness`, `--profile`, `--benchmark`, `--full-benchmark` modes)
        - COMMANDMENT format rules (the harness will be referenced in COMMANDMENT.md)
-       Follow these rules when creating or adapting the test harness.
-       If `src/minisweagent/run/preprocess/INSTRUCTIONS.md` does not exist, follow the rules in this system prompt instead.
+       If that exact path is not directly readable, follow the rules in this system prompt instead.
     5) Determine the **absolute path** of the repo root (run `pwd` or use task context).
     6) If discovery found high-confidence existing tests or benchmarks, **read and validate them first**:
        - If a benchmark script exists (e.g. `benchmarks/benchmark_*.py`), read it — it shows the correct import pattern and shapes.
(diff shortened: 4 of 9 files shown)


Umangatamd:
Implemented a prompt/orchestration fix for preprocess harness repair.

What changed:
- strengthened the UTA prompt and shared preprocess instructions to preserve the full source execution contract, not just shapes
- expanded the shape fixer from shape-only checking into minimal harness repair, including focused validation commands and stderr-aware diagnosis
- added a feedback-driven retry in preprocess: if post-fix revalidation fails, rerun the shape fixer once with the failing mode error before reverting
- added targeted unit coverage for the new prompt guidance and shape-fixer task construction

What this fixed:
- addressed the rotary preprocess failure mode where the harness matched sampled cases but drifted the source tensor contract and later failed at runtime with `ValueError: cos_sin_cache should be float32`
- more generally, this fixes prompt-level harness drift that shape-only checking could miss

Tested:
- `PYTHONPATH=src pytest tests/run/test_preprocess_unit.py -q --noconftest`
- preprocess-only run inside `geak-agent-upandey` for `/sgl-workspace/sglang/sgl-kernel/csrc/elementwise/pos_enc.hip`
- observed all harness modes pass, shape-fixer revalidation pass, profiling complete, baseline metrics saved, and `COMMANDMENT.md` generated
- confirmed the previous `cos_sin_cache` failure no longer appears in that rotary run

iraj465 (approved):

sdubagun-amd (approved):

Issue #144, opened 2026-04-13
Failed to detect GPU info for device 0: 'DeviceSpecs' object has no attribute 'l2_bandwidth_gbs'


zjin-lcf:
### Describe the issue

Please advise if naive users could ignore or need to verify the attributes that failed to be detected

ERROR    [METRIX] rocprofv3 failed with exit code 2

Umangatamd:
Hi Zheming, thanks a lot for pointing out the issues. If possible, can you share the command, setup and the logs. Thanks a lot

zjin-lcf:
System: 4-MI300A  

rocprofv3 --version
             version: 1.1.0
        git_revision: fc0010cf6a5a972d42b276df946510f30343d493
        library_arch: x86_64-linux-gnu
         system_name: Linux
    system_processor: x86_64
      system_version: 6.8.0-71-generic
         compiler_id: GNU
    compiler_version: 13.3.0
        rocm_version: 7.2.0

Command:
geak  --kernel-url '[https://github.com/ROCm/aiter/blob/main/aiter/ops/triton/topk.py'](https://github.com/ROCm/aiter/blob/main/aiter/ops/triton/topk.py%27)   --gpu-ids '0,1,2,3' --yolo   --task 'Optimize the topk kernel.' --exit-immediately   -o '$HOME/workspace/GEAK_ARTIFACTS/topk_wo_task_wo_harness'   2>&1 | tee '$HOME/workspace/GEAK_ARTIFACTS/topk_wo_task_wo_harness.log'


The log:
minisweagent.run.preprocess.preprocessor: INFO: --- Step 5/7: Kernel profiling (Metrix instrumented) ---
[04/13/26 16:47:46] INFO     Profiler MCP: backend=metrix, command=python                                                    server.py:246
                             /home/user/GEAK/$HOME/workspace/GEAK_ARTIFACTS/topk_wo_task_wo_harness/.geak_resolved/ROCm_aiter
                             /main-e15b75d464bf/op_tests/triton_tests/test_topk.py --profile
                    INFO     Command contains shell constructs, wrapping in bash -c: python                                   server.py:62
                             /home/user/GEAK/$HOME/workspace/GEAK_ARTIFACTS/topk_wo_task_wo_harness/.geak_resolved/ROCm_aiter/
                             main-e15b75d464bf
                    INFO     Warmup run 1/2: bash -c 'python                                                                 server.py:190
[...]

PR #164, opened 2026-04-17
fix(metrix-mcp): tolerate DeviceSpecs API drift for GPU info

parth3690:
## Summary

Metrix `DeviceSpecs` no longer guarantees `l2_bandwidth_gbs` (and other optional fields may rename across releases). The MCP server previously accessed these attributes directly, which raised `AttributeError` and caused GPU detection to fall back to `detected: false` even when core identity fields were present.

## Changes

- Add `_pick_spec_field` / `_device_specs_to_gpu_info_dict` to map `device_specs` defensively with `getattr`-style resolution and a small set of attribute aliases (including peak L2 bandwidth).
- `_get_gpu_info_from_metrix` now uses the mapper so missing optional metrics do not invalidate the whole GPU info block.

## Related

- Addresses https://github.com/AMD-AGI/GEAK/issues/144 (the `DeviceSpecs` / `l2_bandwidth_gbs` failure). Note: `rocprofv3` exit errors reported in the same thread are separate from this mapping bug.

## Testing

- `ruff check mcp_tools/metrix-mcp/src/metrix_mcp/core.py`


zjin-lcf:
The three names indicate the same meaning: peak L2 bandwidth. Is that right ?
 
 "l2_bandwidth_gbs",    "peak_l2_bandwidth_gbs",        "l2_peak_bandwidth_gbs"

PR #90, opened 2026-03-31
  feat: integrate RAG MCP knowledge base retrieval into main

Puyuan-Yang:
## Summary
#139 
  Summary                                                                                                                                                                                                             
                                                                                                                                                                                                                      
  Add a standalone RAG MCP server (rag-mcp) to provide GPU/ROCm/HIP optimization knowledge to all agent types (Default, Strategy, Heterogeneous) via MCP tool integration.                                            
                                                                                                                                                                                                                      
  Key Changes                                                                                                                                                                                                         
                                                                                                                                                                                                                      
  - New rag-mcp MCP server (mcp_tools/rag-mcp/): A self-contained MCP service exposing two tools — query (topic-based knowledge search) and optimize (kernel-type + GPU-model optimization suggestions). Uses hybrid  
  retrieval: FAISS embedding search + BM25 keyword search + RRF fusion + BGE reranker.
  - RAG postprocessor (src/minisweagent/tools/rag_postprocessor.py): An LLM-based postprocessor that filters, deduplicates, and restructures RAG retrieval results before passing them to the downstream agent.       
  - Unified RAG config toggle: Single tools.rag: true/false flag in geak.yaml controls RAG availability across all agent types. Startup fail-fast checks verify that the rag-mcp package is installed and the semantic
   index has been built.                                                                                                                                                                                              
  - Dynamic tool prompting: Injects available tool names into agent prompts via tool_names template variable so agents are aware of RAG tools when enabled.                                                           
  - Knowledge base relocation: Moved knowledge-base/ into mcp_tools/rag-mcp/knowledge-base/ to co-locate it with the RAG server. Removed stale customer kernel report markdown files.                                 
  - Heterogeneous orchestrator support: RAG tools (query, optimize) are now available in the orchestrator's tool set, task generator read-only tools, and schema definitions.                                         
  - Legacy RAG code cleanup: Removed unused langchain_retrieval.py, mcp_environment.py, prompts.py, and run_agent.py from mcp_integration/.                                                                           
                                                                                                                                                                                                                      
  Test Plan                                                                                                                                                                                                           
                                                                                                                                                                                                                      
  - Verify pip install -e mcp_tools/rag-mcp installs successfully                                                                                                                                                     
  - Run python scripts/build_index.py --force to build the semantic index
  - Run with rag: true — confirm query and optimize tools are available and return results                                                                                                                            
  - Run with rag: false — confirm RAG tools are disabled and agent runs without errors                                                                                                                                
[...]

PR #166, opened 2026-04-19
feat(memory): enhance cross-session memory retrieval + RAG

iraj465:
## Summary

Integrates the [rag-mcp](https://github.com/AMD-AGI/GEAK/pull/90) server with GEAK's cross-session memory system and expands the shared experience knowledge base. These two paths now have clearly non-overlapping roles and work together to give the agent richer, more portable context without prescribing decisions.

### 1. Cross-session memory KB (`src/minisweagent/memory/cross_session/`)

Per-kernel verified experiences — injected at task start so the agent has concrete past evidence to cross-reference against its current kernel, profile, and hardware.

Each experience carries 29 schema fields populated with real data (no nulls, no empty arrays), including:

- `original_kernel_code` — full source of the kernel at optimization time (enables code-identity matching)
- `patch_content` — real git-style diff that produced the verified result
- `baseline_latency_ms` / `best_latency_ms` / `bottleneck_type` / `hardware`
- `kernel_structure` / `kernel_category` / `kernel_language` / `kernel_url`
- `best_strategy` / `best_change_category` / `code_changes_summary` / `key_insight`
- `strategies[]` — per-round attempts (winners + regressions) with measured outcomes
- `round_insights[]` / `trajectory_sketch` — what worked, what failed, why
- `profiling_insight` / `profiling_metrics` / `top_kernels` / `baseline_benchmark`
- `source_logs` / `verified_speedup_source` / `timestamp` / `record_id` / `success`

### 2. rag-mcp `query` / `optimize` tools (`mcp_tools/rag-mcp/`)

Generic GPU / ROCm / HIP background knowledge (AMD aiter customer-case reports, optimization guides, kernel-type best practices). **Pulled on demand** by the agent when it needs to consult domain reading — not always-injected. Uses FAISS + BGE-large (1024-dim) + BM25 hybrid retrieval with a BGE reranker.

One-time setup per container:
```
pip install -e mcp_tools/rag-mcp
python scripts/build_index.py --force
```

### Retrieval (`retriever.py`)
- Name-stem overlap boost: lifts same-family seeds above noisy same-language matches
- Scaled success boost: higher-impact KB entries rise to the top of retrieval
- Threads `target_kernel_path` through stage-2 scoring and the formatter for code-aware context
- Unified speedup threshold: reads `GEAK_MEMORY_MIN_SPEEDUP` once from config and applies it consistently across retriever boost gates, formatter categorization, and KB-write filter
- Integrates with the rag-mcp tool path (generic RAG is no longer injected in-process; the agent calls it on demand)

### Formatter (`formatter.py`)
- Uniform per-entry evidence block: hardware, baseline → best latency, bottleneck, `code_fingerprint`, kernel structure, key insight, change summary, numerical profiling metrics, strategies with full diffs (both winners and regressions)
- `_code_fingerprint` — short, deterministic `Nbytes, sha256=...` identity so the agent can byte-compare its own kernel against KB entries and decide verbatim vs translate
- `_extract_key_params` surfaces `num_warps`, `BLOCK_SIZE`, `dtype_cast`, etc. pulled from winning patch diffs as a short bullet list
- "Informed cross-reference, not directive" framing: the KB exposes raw evidence; the agent decides whether to adopt, adapt, or set aside. Earlier prescriptive banners and special-case shortcuts removed
- Balanced context budget + truncation to surface enough signal without diluting prompt bandwidth

### Experience record + extractor (`schemas.py`, `extractor.py`)
- `ExperienceRecord` now stores `original_kernel_code` and `strategies[]`
- Extractor auto-populates all 29 schema fields at end-of-run when `GEAK_SAVE_TO_KNOWLEDGE_BASE=1` — `kernel_url`, `profiling_insight`, `baseline_benchmark`, `kernel_structure`, `round_insights`, `strategies` with full diffs, etc. — so auto-saved entries match the field-completeness of curated seeds

### Patch-apply (`generated_artifacts.py`, `save_and_test.py`, `evaluation.py`)
- Fallback-only `git apply --3way` with `object_alternates` registered from the per-round worktree object stores, so eval worktrees can apply patches whose parent blob lives in a sibling sub-agent's bootstrap commit
- `normalize_patch_paths` fixes `diff -ruN` absolute-path headers so patches apply cleanly in the eval worktree
- Empty-patch overwrite guard on `save_and_test`: never destroys a previously-captured non-empty patch with a 0-byte capture
- Explicit all-caps warning surfaced to the agent when `save_and_test` detects no code changes — tells the agent the benchmark ran on baseline and what to do

## Knowledge base expansion (`knowledge_base.json`)

Repopulated with verified experiences accumulated through recent runs:

- **23 entries covering 13 unique kernels** across normalization, RoPE, quantization, GEMM, MoE, point-cloud, and sorting families: `ball_query`, `fused_mxfp4_quant_moe_sort`, `fused_qk_rope_cache_mla`, `fused_qkv_rope`, `fused_rms_fp8`, `gather_points`, `gemm`, `gemm_a16w16_atomic`, `knn`, `llama_ff_triton`, `moe_routing_sigmoid_top1`, `three_nn`, `topk`
- Every entry carries a real patch body (3 KB - 38 KB) and the full `original_kernel_code` (2 KB - 48 KB)
- All 29 schema fields populated (0 empty cells across all entries)
- Every entry passes the `GEAK_MEMORY_MIN_SPEEDUP` verified threshold — no speculative/hypothesized strategies
- Earlier blueprint-only / speculative entries purged; remaining set is ready for cross-kernel technique transfer

## Utility scripts

- `scripts/inject_offline_seed.py` — parses standalone geak logs + takes an explicit `--kernel-source-file` argument, extracting all fields for KB seeding (no reliance on path conventions)
- `scripts/inject_l3_winners.py` — batch-injects discovered winner strategies with full source capture
- `scripts/dedupe_kb.py` — deduplicates by `(kernel_name, best_strategy)` while keeping the highest verified entry
- `scripts/compare_mem_delta.py` — generates mem-off / mem-on comparison tables from canonical snapshots

## Also includes patch-apply fixes (from PR #161 that this branch stacks on)

- `fix(postprocess): fallback to 3-way apply when eval worktree lacks parent blob`
- `fix(save_and_test): prevent empty patch from overwriting non-empty capture`
- `fix(patch-apply): normalize diff -ruN absolute paths to git-style relative`
- `fix(save_and_test): surface explicit warning when no code changes detected`

## Design principles

- **Additive only**: all changes are opt-in. When no KB entry passes the relevance gate, behavior is identical to before. Relevance gate still requires `best_text_sim >= 0.05 OR category match` before injecting memory.
- **Informed cross-reference, not directive**: the formatter exposes raw evidence (baseline code, strategies, diffs, profiling metrics, trajectories) and lets the agent decide. Same contract for every entry, no special-case hand-holding.
- **Portable knowledge**: identity is established by code (`original_kernel_code` + `code_fingerprint`), not by names or paths. The KB is shareable across machines, users, and kernel renames.
- **Consistent threshold**: `GEAK_MEMORY_MIN_SPEEDUP` is a single config value used everywhere (KB-write filter, retriever boost gate, formatter categorization).

## Test plan

- `GEAK_USE_KNOWLEDGE_BASE=1` — retrieval fires, formatter emits per-entry evidence block including agent's own fingerprint + each entry's fingerprint for byte comparison
- `GEAK_MEMORY_DISABLE=1` — no memory injected (ablation)
- `tools.rag: true` in `geak.yaml` — rag-mcp `query` / `optimize` tools are available for the agent to call on demand; `tools.rag: false` disables them cleanly
- Extractor auto-save with `GEAK_SAVE_TO_KNOWLEDGE_BASE=1` on a completed run produces a KB entry with all 29 schema fields populated (no nulls, no empties)
- `scripts/dedupe_kb.py` removes duplicates without data loss
- `scripts/inject_offline_seed.py --kernel-source-file <path>` correctly parses a geak log and appends a fully-populated entry

Made with Cursor


src/minisweagent/tools/save_and_test.py
```diff
@@ -23,6 +23,85 @@
 logger = logging.getLogger(__name__)
 
 
+def _normalize_diff_ruN_to_git(patch_text: str, base_repo: Path, cwd: Path) -> str:
```
yueliu14, line 26: I didn't see where we used diff -runN for patch saving. Please give me a hint.

yueliu14 (changes requested):

src/minisweagent/tools/save_and_test.py
iraj465: `diff -ruN` is produced inside the same file, in the fallback branch of `_get_patch_content` (lines ~699-725). The flow is:

1. **Primary path** (lines 674-697): if the sub-agent's `cwd` IS a git repo, use `git add -N . && git diff --binary` — produces standard git-style patches, no normalization needed.

2. **Fallback path** (lines 699-725): if `cwd` is NOT a git repo but `ctx.base_repo_path` exists, shell out to:
   ```python
   subprocess.run(["diff", "-ruN", "--exclude=.git", "--exclude=__pycache__",
                   *extra_excludes, str(ctx.base_repo_path), str(cwd)], ...)
   ```
   This produces headers with absolute paths:
   ```
   diff -ruN /abs/base/kernel.py /abs/cwd/kernel.py
   --- /abs/base/kernel.py    2026-04-17 19:10:02 +0000
   +++ /abs/cwd/kernel.py     2026-04-19 01:21:00 +0000
   ```
   `git apply` in the eval worktree rejects these with `error: kernel.py: No such file or directory` because the `/abs/base/...` path doesn't exist there. `_normalize_diff_ruN_to_git` rewrites these headers to git-style `a/kernel.py / b/kernel.py` at line 725 right after the `subprocess.run` call.

**When does the fallback fire?** When a parallel sub-agent's worktree is a plain directory copy rather than a proper git worktree — e.g. when `ParallelAgent` sets `base_repo_path = repo_path_resolved` (parallel_agent.py:501) but the per-slot worktree was created as a file copy without `.git`. Without the normalization, those patches look fine at capture time but silently fail at FULL_BENCHMARK-apply time, causing the "captured-but-not-verified" failure mode described in the commit message (`dd3fdd7`).

The second-layer safety (a retry-with-normalization in `apply_patch_with_generated_helper_fallback`) catches any such patches already in flight from before this fix landed.

yueliu14 (approved):

Umangatamd (approved):

PR #178, opened 2026-04-22
 prefer rocprofv3 --kernel-trace over triton.testing.do_bench for kernel benchmark

rbrugaro-amd:
I run GEAK over **batched_gemm_a16wfp4** it claimed a 3x perf improvement and provided a new config json.
I did rocprofv3 kernel profiling on the new  kernel config in isolation and there is actually regression so the uplift came from other optimizations not kernel itselg. 

The issue i believe is that GEAK used triton.testing.do_bench when measuring performance and that measures wall-clock time and was misguided to believe it was a kernel uplift itself. 

This PR contributes a few changes to the agent and instruction to make it aware that triton.testing.do_bench is not appropriate for kernel only evaluations.  
Screenshot 2026-04-20 232903

I created this draft PR in aiter with all the geak results and a summary of the optimizations for context: 
image


src/minisweagent/run/preprocess/INSTRUCTIONS.md
```diff
@@ -210,6 +210,45 @@ script that imports the kernel, creates test inputs, and provides
...
+    # Timed iterations
+    latencies = []
+    for _ in range(ITERATIONS):
+        start = torch.cuda.Event(enable_timing=True)
+        end = torch.cuda.Event(enable_timing=True)
+        start.record()
+        kernel_fn(...)
+        end.record()
+        end.synchronize()
+        latencies.append(start.elapsed_time(end))
```
fxmarty-amd: You need to flush L2 cache between different iterations. AFAIK it is something that `do_bench` handles. See: https://github.com/triton-lang/triton/blob/63e38fe514711419810a54ad9ba5d9de856387e8/python/triton/testing.py#L181-L182

I am actually surprised by your observation. Isn't https://github.com/triton-lang/triton/blob/63e38fe514711419810a54ad9ba5d9de856387e8/python/triton/testing.py#L174-L186 the same as you suggest?

rbrugaro-amd: @fxmarty-amd  you are correct, what i had included in the INSTRUCTION was equivalent with same issue as the do_bench. Actually what I did to measure the before and after the GEAK optimized kernel to realize there was no uplift was to use `rocprofv3 --kernel=trace.` I've done a new commit to reflect that.

src/minisweagent/run/preprocess/INSTRUCTIONS.md
```diff
@@ -210,6 +210,31 @@ script that imports the kernel, creates test inputs, and provides
...
+    **When dispatch-inclusive timing becomes a problem:**
+    For optimization strategies that only change kernel configs (tile sizes,
+    `num_warps`, `num_stages`, etc.), Python dispatch overhead is constant
+    across configs.  If that overhead dominates kernel execution time — common
+    with small, launch-latency-bound shapes — then `do_bench` / CUDA-event
+    timers cannot distinguish between configs, and GEAK may credit unrelated
+    wrapper changes (e.g. `deepcopy` → `dict()`) as kernel speedups.
+
+    **Use `rocprofv3 --kernel-trace` for kernel-only timing:**
+    The harness `--profile` mode already runs the kernel for `rocprofv3`
```
sdubagun-amd, line 231: I don't think the harness in profile mode is implemented this way consistently across the preprocessing components. cc: @Umangatamd.

src/minisweagent/run/preprocess/INSTRUCTIONS.md
```diff
@@ -210,6 +210,31 @@ script that imports the kernel, creates test inputs, and provides
...
+    capture.  When the optimization target is pure GPU kernel performance
+    (config tuning, algorithmic changes), prefer `kernel-profile` /
```
sdubagun-amd, line 233: The reason we kept the benchmark mode separate from the profile mode is because the latter takes a longer execution time. Due to this reason, we also allow the profiling-specific input shape set to be smaller than that of the regular benchmarking set. Due to these design choices, we do not use the latency value from the profiling run, but instead use the value from the benchmarking set. We, of course, use the rest of the parameters from profiling as a "guidance" for optimization.

sdubagun-amd (commented):
The branch seems to be behind GEAK by 100+ commits. Would you be able to rebase to the latest main? Alternatively, could you at least manually make sure this is not a concern?

rbrugaro-amd:
@sdubagun-amd the last commit before my changes was 6 days ago (81c87bd)  and I see the last commit in INSTRUCTIONS.md was 2 weeks ago. But i have merged main now and will check next time.
If you see here: https://github.com/ROCm/aiter/pull/2822 GEAK produced two optimized configs
aiter/ops/triton/configs/gemm/gfx950-BATCHED_GEMM_PREQUANT-AFP4WFP4-N=128-K=512.json
aiter/ops/triton/configs/gemm/gfx950-BATCHED_GEMM_PREQUANT-AFP4WFP4-N=512-K=128.json
which when I microbench with rocprofv3 against the baseline and saw a 5% regression. 

My point with the suggested changes is that during **kernel** optimization/config tuning we should be using rocprofv3 --kernel-trace instead of triton.do_bench (wall clock time) to avoid having GEAK assume is a kernel improvement when the measured improvement may come from reducing some other overhead. If the concern is the time it takes to run during the iterative optimization process at least we can ask it to verify it before delivering a new config?

iraj465:
@rbrugaro-amd — your observation is exactly right and we've since validated it with a systematic ablation. Running the full GEAK loop with `do_bench` vs `torch.profiler` CUDA vs `rocprofv3` as the optimization objective, we found:

- **`do_bench` produced a kernel regression** (0.91x by rocprofv3 ground truth) while reporting 1.13x speedup — the agent optimized dispatch overhead, not the kernel
- **`rocprofv3` in-loop was too slow** for iterative optimization — subprocess overhead per shape made the agent waste rounds
- **`torch.profiler` CUDA** gives kernel-only timing like rocprofv3 but in-process and 240x faster, producing the best true kernel speedup (1.06x, verified by cold rocprofv3)

This is addressed in [PR #208](https://github.com/AMD-AGI/GEAK/pull/208) (`--target=kernel`), which:

1. Makes the UTA emit harnesses with both `wall_ms` and `kernel_ms` signals
2. Uses `torch.profiler` CUDA as the kernel timing method (fast enough for iterative use)
3. Cross-verifies final candidates with cold-subprocess rocprofv3 via the speedup cap

Your PR's core insight is correct but the specific approach (rocprofv3 in the inner loop) is addressed more efficiently in [#208](https://github.com/AMD-AGI/GEAK/pull/208). Happy to coordinate if there are other aspects from your PR we should incorporate.

Note: Since GEAK is an agent like other OOB ones, we still have use-cases where optimizing the whole wall-clock times is required. To address your concerns, we have added specific modes, where GEAK either targets (wall-clock time optimization) or kernel (GPU) only timing optimizing with the `--target` flag - It would be great if you can try that out, also using more GPUs x rounds helps getting more aggressive kernel optimizations.

PR #244, opened 2026-05-26
fix(save_and_test): fall through to diff -ruN when git diff fails or is empty

### LINKED ISSUE #250: GEAK in hyperloom fails to apply patches
### Describe the bug

GEAK fails to apply patches in Hyperloom when `git diff` fails due to:

1. Untracked files or partially copied files due to the existence of submodules.
2. JIT cache sections that get cleared up and render patches invalid.

sdubagun-amd:
## Two complementary fixes to `_get_patch_content`

This PR composes two orthogonal fixes against the same function in `src/minisweagent/tools/save_and_test.py`. Both are needed for the next 24h GEAK session; they were authored independently and now coexist on this branch.

### Fix A — fall-through when `git diff` fails or is empty (this branch's original change)

Slot worktrees created by `git worktree add` (without `--recurse-submodules`) leave submodule `.git` files missing under `3rdparty/`. The first command in `_get_patch_content`'s git branch:

```
git add -N . && git diff -- . ...
```

then aborts with:

```
fatal: not a git repository: '3rdparty/composable_kernel/.git'
```

…returns no stdout, exits non-zero, but the previous code did `return result.stdout` regardless. The orchestrator recorded a zero-byte `patch_*.patch`, Hyperloom clamped the round to `1.00x` speedup, and the round was **discarded as "no changes detected" — even when the agent really did edit `kernel.py` and the test really did pass.**

Reproduction fingerprint from the 20260523 GEAK session: `geak-89318a6d/results/round_5/fixed-canonical-fill-1/patch_0.patch` is exactly 0 bytes despite a passing test on a non-trivial agent edit.

Two minimal changes inside `_get_patch_content`:

1. **Add `3rdparty/` (git pathspec form) and `3rdparty` (diff exclude form)** to the existing exclude lists. Keeps the common case from tripping the submodule traversal in the first place.
2. **Fall through from the git branch to the `diff -ruN` backup branch when `git diff` returns non-zero or empty stdout.** `diff -ruN` does not understand submodule pointers and is robust to this failure mode.

The fall-through is a strict superset of current behaviour:
- `git diff` succeeds with non-empty patch → unchanged.
- `git diff` fails or empty AND `base_repo_path` is set → previously returned `""`; now returns the `diff -ruN` patch.
- `git diff` fails or empty AND no `base_repo_path` → unchanged.

### Fix B — drop JIT-cache sections from rendered patches (merged in from `fix/final_report_optimized_codes`, originally authored by @chao-xu-spec)

`git add -N . && git diff` can capture JIT runtime caches (e.g. `flydsl_cache/*.pkl`) when a harness imports the package. Those pkls leak into the patch text, get persisted into `final_report.optimized_codes`, and pollute downstream analysis.

This PR includes the original author's two-layer fix verbatim:
- **Layer A** — `_strip_jit_cache_from_patch` post-processes the rendered patch text in `_get_patch_content` to drop any `diff --git` section whose path is a JIT cache.
- **Layer B** — `run/postprocess/optimized_codes.py` provides a defence-in-depth pass on the final-report side, plus wiring into `pipeline_types.py`, `results.py`, `homogeneous_agent.py`, and a new public helper `strip_jit_cache_sections` in `run/utils/generated_artifacts.py`.

### Conflict resolution

The two fixes both rewrote the single `return result.stdout` line on the git-branch success path. Resolved combined form:

```python
if result.returncode == 0 and result.stdout.strip():
    return self._strip_jit_cache_from_patch(result.stdout)
# ... fall through to diff -ruN backup
```

The `diff -ruN` branch auto-merged cleanly: `_strip_jit_cache_from_patch(_normalize_diff_ruN_to_git(...))`.

## Tests

29 tests pass across the merged surface:

- **6 fall-through tests** in `tests/tools/test_save_and_test_diff_fallthrough.py` — happy path, both fall-through triggers, no-fallback degenerate case, and `3rdparty` exclude pinning in both branches.
- **1 composition test** — `test_git_branch_success_path_still_strips_jit_cache` — pins that both fixes coexist on the git-branch success path: a captured patch containing a real `kernel.py` edit AND a `flydsl_cache/*.pkl` section keeps the kernel edit and drops the JIT-cache section.
- **17 tests** in `tests/run/test_optimized_codes.py` (Fix B's full suite, unchanged).
- **5 existing** `tests/tools/test_save_and_test_registry.py` tests (regression-clean).

`ruff check` and `ruff format --check` are clean on all modified files.

## Risk

Lowest tier on Fix A side (behaviour unchanged on every existing successful path). Fix B side has its own regression coverage (17 tests).

### DIFF
--- a/src/minisweagent/tools/save_and_test.py
+++ b/src/minisweagent/tools/save_and_test.py
@@ -765,6 +765,14 @@ class SaveAndTestTool:
                 # Build / compile caches
                 "__hip_fatbin/",
                 ".cache/",
+                # Vendored submodules — when slot worktrees are created via
+                # ``git worktree add`` without ``--recurse-submodules``, the
+                # submodule's own ``.git`` is missing under ``3rdparty/``.
+                # ``git diff`` then aborts with ``fatal: not a git repository:
+                # '3rdparty/<sub>/.git'``, returns no stdout, and the caller
+                # records an empty patch. Excluding the path keeps git diff
+                # scoped to files the agent is actually allowed to edit.
+                "3rdparty/",
                 *self._generated_helper_excludes(),
             ]
             exclude_args = " ".join(f"':(exclude){entry}'" for entry in excludes)
@@ -776,7 +784,21 @@ class SaveAndTestTool:
                 timeout=30,
                 shell=True,
             )
-            return result.stdout
+            if result.returncode == 0 and result.stdout.strip():
+                return result.stdout
+            # Fall through to the diff -ruN backup branch below. Reasons we
+            # land here in practice:
+            #   * git diff aborted with non-zero exit (e.g. an uninitialised
+            #     submodule's ``.git`` is missing inside the worktree, even
+            #     after the explicit ``3rdparty/`` exclude above — git still
+            #     traverses submodule pointers before applying pathspecs in
+            #     some configurations);
+            #   * git diff returned cleanly but stdout was empty AND a
+            #     ``base_repo_path`` is available, in which case ``diff -ruN``
+            #     is a strict superset (it picks up genuinely-new files that
+            #     ``git add -N`` may have failed to register).
+            # If ``base_repo_path`` is not available we drop to the final
+            # ``return ""`` below, preserving today's behaviour for that case.
 
         if ctx.base_repo_path and ctx.base_repo_path.exists():
             excludes = [
@@ -800,6 +822,11 @@ class SaveAndTestTool:
                 "*.ncu-rep",
                 "__hip_fatbin",
                 ".cache",
+                # See git-branch comment above: vendored submodules confuse
+                # diff-on-worktrees the same way they confuse git diff. Pass
+                # the bare directory name (no trailing slash) because that's
+                # the form ``diff --exclude=`` accepts.
+                "3rdparty",
                 *self._generated_helper_excludes(),
             ]
             if ctx.patch_output_dir:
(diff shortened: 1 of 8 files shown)


src/minisweagent/tools/save_and_test.py
```diff
@@ -800,6 +826,19 @@ def _get_patch_content(self) -> str:
                 "*.ncu-rep",
                 "__hip_fatbin",
                 ".cache",
+                # See git-branch comment above: vendored submodules confuse
+                # diff-on-worktrees the same way they confuse git diff. Pass
+                # the bare directory name (no trailing slash) because that's
+                # the form ``diff --exclude=`` accepts.
+                "3rdparty",
```
yueliu14: PR #248 also addresses the submodule issue, but at the worktree-creation layer. It adds                    _demote_submodule_gitlinks() which removes 160000 gitlink entries from the index, re-adds the directories as regular files, and commits a baseline — so by the time _get_patch_content runs, git diff never encounters submodule pointers at all. With that in place, the 3rdparty/ exclude and the fallthrough logic in this PR become redundant. The JIT cache stripping part is independent and still needed.

sdubagun-amd: Okay, since the other PR is still under review, I'd suggest we keep this for now. Once the changes are in, we could revert these specific changes.

Umangatamd (approved):

yueliu14 (approved):

Issue #245, opened 2026-05-27
User provided harness missing GEAK_WORK_DIR


yueliu14:
## Description
When user provide test command with hardcode build path, like /sglang-workspace/aiter/, GEAK will always test the original hardcode path instead of optimized one in worktree. Thus, the speedup is always 1.0

## How to solve the issue
For user-provided test command, GEAK adds a harness modify module to refine the harness file.

PR #248, opened 2026-05-27
Harness Worktree Sanitize

yueliu14:
## Motivation
Fix issue #245 

When users supply custom test commands (e.g. `--test-command "cd /path && bash run_eval.sh"`), the referenced scripts often contain hardcoded absolute paths to the target repo. Since GEAK creates isolated worktrees for each optimization agent, hardcoded paths cause the agent to test against the original repo instead of its own modified copy — silently invalidating all benchmark results.

## Main Feature: Harness Sanitizer

Added a new **harness-sanitizer** subagent (`harness_sanitizer.py`) that automatically:

1. **Sanitizes hardcoded paths** — discovers all scripts referenced by the test command, detects hardcoded repo paths, and rewrites them using GEAK env vars (`GEAK_WORK_DIR`, `GEAK_GPU_DEVICE`, etc.)
2. **Injects missing compile steps** — checks whether scripts include a build step (make, cmake, hipcc, `pip install -e`, etc.) and prepends one if missing, so each agent recompiles from its own modified source tree
3. **Standardizes benchmark output** — detects how results are reported and emits `GEAK_RESULT_METRIC` / `GEAK_RESULT_UNIT` / `GEAK_RESULT_DIRECTION` markers for consistent downstream evaluation

## Enhancements

- **Generalized benchmark metrics**: Introduced `BenchmarkMetric` dataclass with `GEAK_RESULT_METRIC` / `GEAK_RESULT_UNIT` / `GEAK_RESULT_DIRECTION` markers to support both `lower_is_better` (latency) and `higher_is_better` (throughput/bandwidth) metrics. Added `compute_speedup()` that respects metric direction. Legacy `GEAK_RESULT_LATENCY_MS` remains supported.
- **Multi-GPU evaluation**: Changed eval pipeline from `gpu_id: int` to `gpu_ids: list[int]` across `build_eval_env`, `resolve_eval_worktree`, `preflight_commandment_contract`, etc. Multi-GPU kernels (e.g. `cross_device_reduce`) now receive all assigned device IDs.
- **HIP kernel language**: Added `"hip"` as a recognized `kernel_language`. HIP kernels now trigger build steps in SETUP and proper build-info resolution.
- **Path A baseline collection**: `collect_baseline` now accepts `eval_command` for Path A flows where the harness doesn't support `--benchmark`. Evaluation fallback recaptures baselines from COMMANDMENT when preprocessing didn't produce one.
- **Configurable timeouts**: Benchmark, profile, correctness, and codebase-explore timeouts overridable via `GEAK_BENCH_TIMEOUT`, `GEAK_PROFILE_TIMEOUT`, `GEAK_EXPLORE_TIMEOUT`. Profiler-mcp calls wrapped with a timeout guard.
- **Subagent wall timeout**: `PreprocessSubagent` gains a `wall_timeout` parameter to cap total wall-clock time.
- **Knowledge base opt-in**: `GEAK_USE_KNOWLEDGE_BASE` changed from on-by-default to off-by-default (set `=1` to enable).

## Bug Fixes

- **Best-patch tracking**: When no verified speedup is available, agent-reported speedup is now used as fallback for `starting_patch` tracking (previously the patch stayed stuck across rounds).
- **COMMANDMENT dedup**: Identical CORRECTNESS / BENCHMARK sections are no longer duplicated in test commands.
- **Dispatcher padding**: Fixed padding tasks to propagate `agent_name`, `kernel_language`, and `num_gpus` from the base fixed task.
- **TaskPlanner**: `build_pool` now uses `num_parallel` instead of `num_gpus` for worker count decisions.
- **Path A discovery skip**: Case A (user-provided command) now skips unnecessary `run_discovery`, going directly to `commandment_from_user_command`.
- **Benchmark parsing consolidation**: Moved canonical logic to `preprocess.benchmark_parsing` with a backward-compat re-export shim in `postprocess`.

## Test plan

- [x] Verify harness sanitizer rewrites hardcoded paths and injects compile steps
- [x] Test Path A flow with user-provided test command + eval_command baseline
- [x] Verify multi-GPU evaluation with cross-device kernels
- [x] Test higher-is-better metrics (e.g. GB/s) produce correct speedup
- [x] Confirm timeout env var overrides work
- [x] Run existing test suite


src/minisweagent/run/preprocess/benchmark_parsing.py
```diff
@@ -185,40 +182,128 @@ def _universal_latency_fallback(text: str) -> float | None:
...
+    For time-based units this is a direct conversion.  For throughput
+    units (higher-is-better) we return a synthetic inverse
+    (``1000 / value``) so that legacy callers comparing in ms-space
+    still get directionally correct results.
+    """
+    factor = _TIME_UNIT_TO_MS.get(metric.unit.lower())
+    if factor is not None:
+        return metric.value * factor
+    if metric.value > 0:
+        return 1000.0 / metric.value
```
sdubagun-amd: Why is this being inverted in all cases? If it is seconds, then it should be metric.value / 1000.0 right?

sdubagun-amd (changes requested):

src/minisweagent/run/preprocess/benchmark_parsing.py
yueliu14: Thanks for pointing it out! I have fixed the metric.value

sdubagun-amd (approved):

chao-xu-spec (approved):

PR #249, opened 2026-05-29
Make GEAK evaluate worktree (patched) kernels, not the baseline

### LINKED ISSUE #246: Auto-generated harness for C++/HIP header-only kernels hard-codes the source repo path and ignores `GEAK_WORK_DIR`, so every patched-candidate evaluation silently re-runs the original unpatched kernel
## TL;DR

For C++/HIP header-only kernels, the harness builder emits a `harness.py`
whose include paths and build cache are **absolute literals into the
source repo**, not derived from `GEAK_WORK_DIR` / `GEAK_REPO_ROOT`.

When the orchestrator creates an eval worktree, applies a candidate
patch, and runs `COMMANDMENT.md`, the harness still compiles and loads
a `.so` built from the **unpatched** header in the source repo. Verified
speedup is always ≈ 1.0× and `--correctness` always PASSes, regardless
of what the optimizer changed. No error, no warning; every contract
validator passes. The optimizer (and every sub-agent's inner-loop
benchmark) is therefore training on a flat signal.

## Affected scope

- Any C++ / HIP / CUDA **header-only template kernel** where the
  harness compiles a `.so` via `hipcc` / `nvcc` and `#include`s the
  target header.
- Triton / Python kernels that rely on
  `PYTHONPATH=${GEAK_WORK_DIR}:...` shadowing are not affected.

## What COMMANDMENT.md does correctly

The Jinja template
`src/minisweagent/kernel_languages/triton/commandment.j2` renders a
correct setup block — cwd and PYTHONPATH are pointed at the worktree:

```bash
## Setup
export GEAK_WORK_DIR="${GEAK_WORK_DIR:-<SOURCE_REPO>}"
printf '#!/bin/bash\nexport PYTHONPATH=%s:%s:${PYTHONPATH}\nexport HIP_VISIBLE_DEVICES=%s\ncd "%s" && exec python3 "$@"\n' \
  "${GEAK_WORK_DIR}" "${GEAK_REPO_ROOT:-<SOURCE_REPO>}" "${GEAK_GPU_DEVICE:-0}" "${GEAK_WORK_DIR}" \
  > ${GEAK_WORK_DIR}/run.sh && chmod +x ${GEAK_WORK_DIR}/run.sh

## Correctness
${GEAK_WORK_DIR}/run.sh <OPT_LOG_DIR>/harness.py --correctness

## Benchmark
${GEAK_WORK_DIR}/run.sh <OPT_LOG_DIR>/harness.py --benchmark

## Full Benchmark
${GEAK_WORK_DIR}/run.sh <OPT_LOG_DIR>/harness.py --full-benchmark

## Profile
kernel-profile "${GEAK_WORK_DIR}/run.sh <OPT_LOG_DIR>/harness.py --profile" \
  --gpu-devices ${GEAK_GPU_DEVICE:-0} --replays 3 \
  --json -o ${GEAK_WORK_DIR}/profile.json
```

`run.sh` correctly exports
`PYTHONPATH=${GEAK_WORK_DIR}:${GEAK_REPO_ROOT}:...` (worktree first) and
`cd`s into `${GEAK_WORK_DIR}` before `exec`ing python3. The
COMMANDMENT side is not the bug — the harness it invokes does not honor
any of these signals.

## Root cause — exact contents of the generated `harness.py`

Generated by the harness-builder (Layer 5 in
`src/minisweagent/run/preprocess/phases/harness.py`; system prompt in
`subagents/harness-generator/SYSTEM_PROMPT.md`). Four problematic
blocks, copied verbatim from the artifact for an `act_and_mul_kernel`
run on `/sgl-workspace/sglang/sgl-kernel/include/hip/hip_act_and_mul.cuh`:

### Block 1 — `REPO_ROOT` is a hard-coded literal

```python
REPO_ROOT = "/sgl-workspace/sglang"             # <-- should derive from ${GEAK_WORK_DIR}
SGL_KERNEL_INC     = os.path.join(REPO_ROOT, "sgl-kernel", "include")
SGL_KERNEL_INC_HIP = os.path.join(SGL_KERNEL_INC, "hip")
KERNEL_HEADER      = os.path.join(SGL_KERNEL_INC_HIP, "hip_act_and_mul.cuh")

HARNESS_DIR = os.path.dirname(os.path.abspath(__file__))
BUILD_DIR   = os.path.join(HARNESS_DIR, "_harness_build")
os.makedirs(BUILD_DIR, exist_ok=True)
```

`os.environ.get("GEAK_WORK_DIR")` / `os.environ.get("GEAK_REPO_ROOT")`
appear nowhere in the file. `cd "${GEAK_WORK_DIR}"` from `## Setup` has
no effect because every subsequent open uses absolute paths.

### Block 2 — `hipcc -I` points at the source repo

The HIP wrapper source the harness compiles:

```cpp
#include <hip/hip_bf16.h>
#include <hip/hip_runtime.h>
#include "hip/hip_act_and_mul.cuh"           // <-- resolved via -I below
// ... instantiates sgl_hip::activation::act_and_mul_kernel<__hip_bfloat16, silu_act<T>>
```

The build command:

```python
def _build_extension():
    # ... writes _SRC_PATH and a stub utils.h ...
    cmd = [
        "/opt/rocm/bin/hipcc", "-O3", "-std=c++17", "-fPIC", "-shared",
        "-DUSE_ROCM=1", "-D__HIP_PLATFORM_AMD__=1", "--offload-arch=gfx942",
        f"-I{stub_inc_dir}",
        f"-I{SGL_KERNEL_INC}",       # <-- /sgl-workspace/sglang/sgl-kernel/include
        f"-I{SGL_KERNEL_INC_HIP}",   # <-- /sgl-workspace/sglang/sgl-kernel/include/hip
        "-x", "hip",
        _SRC_PATH,
        "-o", _LIB_PATH,
    ]
    subprocess.run(cmd, check=True, capture_output=True, text=True)
```

Because `#include "hip/hip_act_and_mul.cuh"` is resolved exclusively
through these `-I` flags, hipcc never sees the worktree copy of the
header — even though the wrapper `cd`'d into the worktree.

### Block 3 — `.so` is cached in the harness's own directory, shared across all worktrees / rounds / GPUs

```python
HARNESS_DIR = os.path.dirname(os.path.abspath(__file__))
BUILD_DIR   = os.path.join(HARNESS_DIR, "_harness_build")
...
_LIB_PATH = os.path.join(BUILD_DIR, "libact_and_mul_silu.so")
_SRC_PATH = os.path.join(BUILD_DIR, "act_and_mul_silu.hip.cpp")
```

`COMMANDMENT.md` references `harness.py` by absolute path, so
`HARNESS_DIR` and `BUILD_DIR` resolve to the **same directory** for
every worktree, every round, and every GPU worker. Concrete evidence
from this run:

```text
$ ls -la <OPT_LOG_DIR>/_harness_build/
act_and_mul_silu.hip.cpp     (1049 B)
libact_and_mul_silu.so       (24984 B, built once)
stub_include/
```

A single `.so` lives outside every worktree. Concurrent GPU workers
race on the same `.cpp`/`.so`.

### Block 4 — Rebuild mtime check walks the source repo, not the worktree

```python
def _build_extension():
    ...
    need_rebuild = True
    if os.path.exists(_LIB_PATH):
        try:
            so_mtime  = os.path.getmtime(_LIB_PATH)
            dep_mtimes = [os.path.getmtime(_SRC_PATH)]
            for root, _dirs, files in os.walk(SGL_KERNEL_INC_HIP):    # <-- source repo
                for fn in files:
                    if fn.endswith((".cuh", ".h", ".hpp")):
                        dep_mtimes.append(os.path.getmtime(os.path.join(root, fn)))
            need_rebuild = so_mtime < max(dep_mtimes)
        except OSError:
            need_rebuild = True
```

Even if (1) and (2) were fixed to point at the worktree, this mtime
detector still scans `<SOURCE_REPO>/sgl-kernel/include/hip/`, whose
files never change during optimization — `need_rebuild = False`, cached
`.so` is loaded, patch ignored.

## End-to-end actual behaviour

| Pipeline step | Intended | Actual |
|---|---|---|
| `git worktree add` + `git apply` | Patches `<WORKTREE>/.../hip_act_and_mul.cuh` | ✅ same |
| `bash` runs `## Setup` | `run.sh` exec's python3 with cwd=worktree | ✅ same |
| `run.sh harness.py --correctness` | Build & test candidate from worktree | Reads `<SOURCE_REPO>/.../hip_act_and_mul.cuh` |
| `_build_extension()` mtime check | Detect candidate change → rebuild | Source-repo mtimes unchanged → skip rebuild, load old `.so` |
| `mode_correctness()` | Candidate vs PyTorch reference | **Original** kernel vs reference → always PASS |
| `mode_full_benchmark()` | Time candidate | Time **original** kernel |
| `_compute_verified_speedup()` | `baseline_ms / candidate_ms` | ≈ 1.0× ± noise |
| `mode_profile()` → `kernel-profile` | Profile candidate | Profile original; `profile_comparison.json` is meaningless |

Sub-agent inner-loop `--benchmark` calls share the same bug, so the
optimizer's own ranking signal is also flat.

## Why no validator catches it

- `kernel_languages/contract.py::validate_harness` only checks the four
  argparse flags and `GEAK_RESULT_*` stdout markers:

  ```python
  REQUIRED_HARNESS_FLAGS   = ("--correctness", "--benchmark", "--full-benchmark", "--profile")
  REQUIRED_HARNESS_MARKERS = ("GEAK_RESULT_LATENCY_MS", "GEAK_RESULT_SPEEDUP")
  ```

- `HarnessBuilder._validate`
  (`src/minisweagent/pipeline_workers/preprocess/harness_builder.py`)
  checks the same set.
- `preflight_commandment_contract`
  (`src/minisweagent/run/postprocess/evaluation.py`) runs
  `SETUP+CORRECTNESS --iterations 1` on the **unpatched** source repo,
  where the harness obviously works.
- `recapture_commandment_baseline` records the baseline on the
  unpatched repo. Since the candidate-side path also reads from the
  source repo, baseline and candidate outputs are byte-identical →
  `_check_config_mismatch` is silent and `_compute_verified_speedup`
  computes `1.0×` without raising.

Result: a green run with verified speedup ≈ 1× and no `failure_reason`
— indistinguishable from a legitimate "optimizer couldn't beat
baseline" outcome.

## Reproducer

You don't need this specific kernel; any HIP / CUDA header-only kernel
exhibits the bug.

1. Run the GEAK preprocessor against a HIP header-only kernel so
   `<OPT_LOG_DIR>/harness.py` + `<OPT_LOG_DIR>/COMMANDMENT.md` are
   emitted (`discovery.json["kernel"]["type"] == "hip"`).

2. Create a worktree of the source repo and deliberately corrupt the
   kernel inside the worktree only — e.g. in
   `<WORKTREE>/sgl-kernel/include/hip/hip_act_and_mul.cuh`:

   ```diff
    template <typename T>
    __device__ __forceinline__ T silu_act(const T& x) {
   -    float fx = __bfloat162float(x);
   -    float y  = fx / (1.0f + expf(-fx));
   -    return __float2bfloat16(y);
   +    return T(0);  // intentionally broken
    }
   ```

3. Execute the COMMANDMENT against that worktree:

   ```bash
   export GEAK_WORK_DIR=<WORKTREE>
   export GEAK_REPO_ROOT=<SOURCE_REPO>
   export GEAK_GPU_DEVICE=0
   bash -c "$(awk '/^## Setup/{f=1;next} /^## Correctness/{exit} f' COMMANDMENT.md \
                | sed -n '/^```bash/,/^```/p' | sed '/^```/d')"
   bash -c "$(awk '/^## Correctness/{f=1;next} /^## /{f=0} f' COMMANDMENT.md \
                | sed -n '/^```bash/,/^```/p' | sed '/^```/d')"
   ```

   - **Expected**: `--correctness` FAILs (`out == 0` vs `ref == silu·mul`).
   - **Actual**: prints `CORRECTNESS_OK`. The cached `.so` from the
     source-repo build is loaded; the broken header in the worktree is
     never seen.

4. Wipe the cache and repeat to isolate Block 2 from Block 3/4:

   ```bash
   rm -rf <OPT_LOG_DIR>/_harness_build/
   ```

   `--correctness` STILL passes — the rebuild uses
   `-I <SOURCE_REPO>/...` and still skips the worktree.

5. Read-only sanity probe (no patching needed):

   ```bash
   grep -E 'GEAK_WORK_DIR|GEAK_REPO_ROOT' <OPT_LOG_DIR>/harness.py
   ```

   On a buggy harness this prints nothing.

## Impact

- Any HIP / CUDA header-only kernel optimization done with this
  generator measures the original kernel, not the candidate. Reported
  speedups are noise.
- Patches that **break correctness** are not detected — the test runs
  against the unmodified code.

chao-xu-spec:
Close #246 
## Problem
GEAK copies a repo into a worktree, applies the candidate patch, then runs
`harness.py`. In several setups the harness silently imported the BASELINE
kernel instead of the worktree, so every speedup measured baseline-vs-baseline
(~1.00x). Three independent failure modes were found:

1. **aiter (JIT C++/HIP)**: `.so` modules were never rebuilt from worktree
   source (`AITER_REBUILD` unset; `module_aiter_core` on a hardcoded
   rebuild-allowlist), so edits never entered the runtime binary.
2. **sgl-kernel / monorepos**: only the FIRST installable sub-project was
   editable-installed, so `import sgl_kernel` resolved to the site-packages
   wheel — baseline again.
3. **Harness hardcoded `sys.path.insert(0, "/sgl-workspace/sglang/python")`**:
   a literal absolute path at sys.path[0] shadowed both the worktree on
   PYTHONPATH and the editable install (observed in
   `rotary_embedding_kernel_202605290819`).

vLLM adds a fourth constraint: it ships as a wheel-only install (no `setup.py`,
multi-GB `.so`), so neither `git worktree` nor `pip install -e` applies.

### 1. Force JIT kernels to rebuild from worktree source
- `_compile_bootstrap/{__init__,sitecustomize}.py` (new): stdlib-only bootstrap
  auto-loaded via PYTHONPATH; sets `AITER_REBUILD=2` and installs an import
  hook clearing aiter's `rebuilded_list` (covers `module_aiter_core`).
- `run/preprocess/run_harness.py`, `tools/save_and_test.py`,
  `run/preprocess_v3/baseline.py`: inject the bootstrap dir + `AITER_REBUILD`
  into every harness subprocess (setdefault — explicit overrides win).

### 2. Editable-install every installable sub-project in the worktree
- `run/preprocess/worktree_install.py` (new): bounded recursive walk installs
  every `setup.py`/`pyproject.toml` sub-project; two-tier strategy
  (`pip install -e` → `setup_rocm.py develop` fallback); snapshots & restores
  the original wheels.
- `run/mini.py`: `geak --cleanup` restores the original wheel-installed
  packages (best-effort).
- `run/utils/generated_artifacts.py`: keep editable-install build side-effects
  (hipify-modified files, generated metadata) out of captured patches.

### 3. vLLM wheel-only support via shadow worktree
- `kernel_packages/{__init__,profile,shadow_worktree,vllm_profile}.py` (new):
  `PackageProfile` registry; shadow tree copies `.py` (writable) and symlinks
  `.so` (immutable baseline), git-inits for clean `git diff`; vLLM profile
  detects wheel-only layout, skips editable install, injects runtime env.
- `run/task_file.py`: `create_worktree` dispatches to a profile's
  `make_worktree` when one matches.
- `agents/parallel_agent.py`: skip `git init` for shadow-tree profiles (never
  pollute system site-packages).

### 4. Reject harnesses that bypass the worktree (the 1.00x bug)
- `run/preprocess/unit_test_agent.py`: forbid hardcoded
  `sys.path.insert(0, '/abs')` in triton/asm/unknown/cuda guidance (only
  hip/ck had it); steer toward `GEAK_WORK_DIR`.
- `kernel_languages/contract.py`: `find_hardcoded_syspath_inserts` +
  `validate_harness` raises `ContractViolation` → HarnessBuilder regenerates.
- `run/preprocess/harness_utils.py`: same detection → `valid=False`, so
  phases and the Path-A short-circuit drop and regenerate the harness.
- `pipeline_workers/preprocess/harness_builder.py`,
  `run/preprocess_v3/tools.py`: wire stricter validation into Path-A.

### 5. Atomic editor writes
- `tools/editor_tool.py`: write via temp file + `os.replace` so edits never
  corrupt a hard-linked/shared baseline inode.

### 6. Bash tool firewall + per-slot env resolution
- `tools/bash_command.py`: L1 wall-clock timeout + L2 filesystem-scan firewall
  (reject `find /`, NFS roots); expand `$VAR`/`${VAR}` from the injected env so
  `GEAK_WORK_DIR` resolves per parallel slot.
- `subagents/**/SYSTEM_PROMPT.md`, `subagents/_common/search_scope_hint.md`,
  `run/preprocess_v3/subagent.py`: tell subagents the allowed search scope.

### Cleanup
- `run/preprocess/commandment.py`, `run/utils/generated_artifacts.py`: remove
  the superseded `.aiter_jit` / `AITER_JIT_DIR` scratch-dir mechanism (replaced
  by AITER_REBUILD + sitecustomize hook); unify cpp/python SETUP template.

## Tests
- `tests/kernel_packages/` (new): compile bootstrap, profile detection,
  shadow_worktree layout, end-to-end dispatch.
- `tests/run/test_worktree_install.py` (new): recursive editable install.
- `tests/run/test_harness_workdir_and_tolerance.py`: hardcoded-sys.path gate
  (both validators), workdir injection, tolerance cap.
- `tests/tools/test_bash_command_safety.py`: firewall + env expansion.


src/minisweagent/models/litellm_model.py
sdubagun-amd, line 1: Could you remove and rotate these keys immediately?

src/minisweagent/memory/working_memory.py
```diff
@@ -68,8 +68,8 @@ class WorkingMemory:
     insights: list[Insight] = field(default_factory=list)
 
     # Budget
-    max_cost: float = 0.50
-    max_steps: int = 100
+    max_cost: float = 0.0
```
sdubagun-amd, line 71: Would this affect working memory in any way? Why are the costs being set to 0?

src/minisweagent/tools/bash_command.py
```diff
@@ -16,6 +18,325 @@
...
+    symlinks (``realpath``) because the caller only needs ancestry, and
+    symlink resolution would be a syscall per check.
+
+    ``extra_env`` is consulted before ``os.environ`` for ``$VAR`` /
+    ``${VAR}`` substitutions.  ``~`` still uses ``os.environ['HOME']``
+    via :func:`os.path.expanduser`; we don't override that because the
+    parallel-agent flow doesn't differ on HOME and the bug we're fixing
+    is specifically about ``GEAK_*``.
+    """
+    expanded = _expand_env_vars(os.path.expanduser(tok), extra_env)
```
sdubagun-amd, line 163: Could you please address the CI issues?

src/minisweagent/kernel_languages/contract.py
```diff
@@ -51,6 +51,105 @@ class ContractViolation(RuntimeError):
...
+# front of ``sys.path`` (index 0) shadows BOTH the GEAK worktree that the
+# COMMANDMENT SETUP prepends to ``PYTHONPATH`` AND the editable install
+# GEAK re-points at the worktree.  The harness then imports the BASELINE
+# package regardless of the agent's edits, so every optimization round
+# measures baseline-vs-baseline (~1.00x) — a silent false-negative.
+# Observed in rotary_embedding_kernel_202605290819 where
+# ``sys.path.insert(0, "/sgl-workspace/sglang/python")`` pinned every run
+# to the baseline sglang checkout.
+_HARDCODED_SYSPATH_RE = re.compile(
+    r"""sys\.path\.insert\(\s*\d+\s*,\s*(['"])(?P<path>/[^'"]*)\1\s*\)"""
```
sdubagun-amd, line 117: `sys.path.append` or `extend` wouldn't work with this.

sdubagun-amd (commented):
Here are some comments.

sdubagun-amd (changes requested):

PR #295, opened 2026-06-24
geak_v4 and hyperloom integration

chao-xu-spec:
# Integrate Hyperloom-driven E2E into GEAK_v4 (alignment, handoff resilience, guaranteed A/B completion)

## Summary
This PR fast-forwards `GEAK_v4` to `fix/geak_v4_hyperloom` (10 commits, +2954/-109 across 16 files, no conflicts). It makes the GEAK-e2e PerfSkills pipeline robustly drivable by an external orchestrator (Hyperloom) while keeping the standalone path unchanged. The work hardens three fragile seams that previously discarded real, measured wins: the caller↔workflow handoff, the integrate A/B measurement, and the newer Claude Code SDK/CLI background-task behavior. It also lands the int4 W4A16 fused-MoE tuning knowledge (the memory-free MoE lever) validated on Kimi-K2.6.

## Motivation
- The workflow return scraped from the agent transcript was a single point of failure: one failed scrape discarded an entire run as `workflow_parse_error` even though every artifact and the measured gain were already on disk.
- An integrate A/B could stop after only the reference leg (timeout/hang) and be silently treated as a rejection — throwing away a real isolated speedup.
- Newer Claude Code builds route `Workflow` to a non-blocking background task, so a one-shot `query()` returned before the pipeline actually finished.
- Measurement mismatch (caller's variable-length `random_range_ratio=0` vs the standalone fixed-length `1` default) introduced a silent ~10–15% gap that made GEAK numbers incomparable to Hyperloom's baseline.

### 1. Hyperloom integration + bench-protocol alignment
- New optional `handoff.bench_protocol` (`random_range_ratio`, `num_prompts`, `num_warmups`, `seed`). `run_e2e.py:apply_bench_protocol` exports ONLY the provided keys to env, which `bench_e2e.sh` inherits. Partial-friendly; omitting it keeps standalone defaults unchanged.
- Integrator role is explicitly forbidden from hard-coding knobs so the caller's protocol always wins; both A/B legs only vary overlay/flags/env.
- New `handoff.phases` / `e2e_repeats` / `state` / `eval_dir` pass-through for phase-scoped resume against a single pinned EVAL_DIR.

### 2. Handoff resilience — the workflow return is never a single point of failure
- Robust capture: SDK path accumulates the full transcript (every text/tool-result fragment), not just the last assistant text.
- Robust extraction: scans the whole transcript for the last JSON object carrying `eval_dir` (compact, ```json``` fenced, pretty-printed, or trailing prose).
- On-disk sentinel: success persists the return to `<eval_dir>/workflow_return.json`.
- Disk recovery: if capture/extraction fails or the run times out after the measured leg, the return is rebuilt from on-disk artifacts (`director_e2e_validation.json`, the stable `overlay/cand_*` layout). Recovered runs set `recovered_from_disk = true`. A real win is never lost.

### 3. Guaranteed A/B completion (three-state gate)
- The integrator MUST measure BOTH reference and candidate legs. New `gate:'incomplete'` + `ab_complete` signals distinguish "didn't finish measuring" from "measured a loss".
- `runIntegrateBothLegs()` re-invokes the integrator (resume mode, reusing the on-disk ref leg) up to `AB_FINISH_RETRIES` times to finish the missing candidate leg.
- Incomplete A/Bs are carried as `pendingIntegrations` (not rejections); Finalize finishes the best pending win's A/B before reporting.

### 4. Claude Code SDK/CLI robustness
- Persistent SDK client + bounded disk grace-poll (`DONE_GRACE_S`/`DONE_POLL_S`) so a backgrounded `Workflow` task / detached Validate leg isn't orphaned when the turn "looks done".
- Inject `enableWorkflows`+`ultracode` settings (public builds reject `--effort ultracode`); `PERFSKILLS_CLAUDE_BIN` to pin the CLI the SDK drives.
- Guaranteed interface-file emission via atexit/signal handlers.

### 5. int4 W4A16 fused-MoE knowledge (the memory-free MoE lever)
- New `knowledge/moe_int4_tuning.md`: per-shape fused-MoE Triton config tune deployed via `VLLM_TUNED_CONFIG_FOLDER` (zero extra HBM); +11–18% e2e (validated +16.4% on Kimi-K2.6, TP8 & TP4).
- Anti-cheat oracle contract for quantized GEMM in `kernel_extractor.md`: candidates see ONLY compact quantized operands (packed int4 + scales), never the dequantized weights or reference output, preventing fake isolated wins.
- `op_benchmarker.md` prioritizes the int4 config tune over an fp8 rewrite (which OOMs at KV-cache init at memory parity).
- New `perf_knowledge/case_studies/by_model/kimi_k2.6_int4_moe_mi300x.md` + INDEX/learned-knowledge updates.

### 6. Infra hardening
- Serving-GPU mutex (`flock`) in `bench_e2e.sh` so profiler/config-sweep/integrate/validation can't starve each other on a shared TP=N serving slot.
- Auto trust-remote-code for the bench client when the server uses it (custom-tokenizer models like Kimi-K2.6), mirrored into the vLLM adapter. Off by default.

## Tests
- New `interface/test_run_e2e_recovery.py`: asserts interface files (`result.json` + `kernel_journey.json`) are always emitted whenever any measured E2E effect exists on disk — no timeout/signal/exception leaves them missing.

## Risk / compatibility
- Fast-forward merge, no conflicts.
- Standalone PerfSkills path is unchanged when `bench_protocol` is omitted; new gate states and recovery are backward-compatible (a legacy integrator that omits `ab_complete` accepts/rejects exactly as before).

## Test plan
- [x] Standalone e2e run (no `bench_protocol`) reproduces prior behavior.
- [x] Hyperloom-driven run forwards `bench_protocol`; A/B matches the caller's baseline.
- [x] Force an incomplete A/B (timeout) and confirm it is carried as pending and finished at Finalize, not rejected.
- [x] Kill the runner after the measured leg and confirm `result.json` is recovered from disk (`recovered_from_disk=true`).
- [x] `python3 -m pytest GEAK/interface/test_run_e2e_recovery.py -v` passes.
- [x] int4 MoE tune produces a `VLLM_TUNED_CONFIG_FOLDER` env winner with parity preserved.


zihaoanllm (approved):
LGTM

PR #306, opened 2026-06-25
feat(kernel_workflow): from-analysis variant workflow

iraj465:
## Summary
Adds `kernel_workflow/kernel_workflow_from_analysis.js`, a GEAK v4 workflow variant that **starts from a pre-computed TraceLens/HL analysis** instead of running v4's own Analyze/Profile phases.

- A new **Seed** phase reads ONLY `analysis_md_path` + `kernel_candidates_path` (the same two files that feed GEAK v3) plus the authoritative `dispatch_prompt_path` (verbatim HL `build_prompt()` output: device source + captured serving shapes + #703 WORKLOAD CONTEXT), and maps them into v4's roadmap — no rocprof / re-trace / re-profile.
- Benchmark (harness build), Optimize, Verify, Merge, Report are byte-identical to stock `kernel_workflow.js`.

## Why
For a controlled GEAK **v3 vs v4** optimizer comparison, both must consume identical info; only the optimization engine should differ. Stock v4 re-derives its own analysis, which breaks apples-to-apples. This variant feeds v4 the exact analysis/candidates/prompt v3 received.

## Usage
Invoke via the Claude Code Workflow tool with args: `kernel_path`, `analysis_md_path`, `kernel_candidates_path`, `dispatch_prompt_path`, `seed_target` (tgNNN), `budget`, `gpu_ids`. Run once per target kernel.

PR #312, opened 2026-06-26
wire upstream TraceLens prior into the GEAK e2e optimization pipeline

chao-xu-spec:
## Summary

Make the GEAK e2e workflow consume the upstream **TraceLens / kernel-agent**
analysis that the orchestrator already produced for the same baseline workload,
instead of re-discovering everything from scratch. The four artifacts are
auto-discovered beside the handoff's `perfskills` dir, forwarded into the
workflow, and used as a prior by the Profile / Strategize / Extract phases.

The change is **entirely additive and null-safe**: when no TraceLens artifacts
are found (or for any post-config reprofile, where the baseline prior is stale)
the workflow profiles/strategizes exactly as before — a tracelens-less run is
byte-identical.

## Motivation

`run_e2e.py` already surfaced a `tracelens_report` block in the driver prompt,
but those paths were **never threaded into the workflow args**, so the JS
phases and role agents could not use them. TraceLens already provides hot-kernel
names, source files, launcher seams, categories, bound types, and shapes — and a
full roofline torch-trace — which overlaps with GEAK's own Profile/Strategize/
Extract work. Reusing it lets GEAK skip an expensive warm-server profiling pass
and start optimization from a richer prior.

## What changed

**`interface/run_e2e.py`**
- `resolve_tracelens_report()` + helpers: glob-discover the four artifacts under
  the experiment root (parent of `perfskills`), with a fallback to the on-disk
  grandparent of the handoff when `exp_root` is not present on the box.
- `map_args()` now injects the **non-null** resolved paths as `args.tracelens`
  (bridge into the JS workflow); the key is omitted entirely when none are found.
- `build_prompt()` still renders a human-readable `tracelens_report` block.

**`e2e_workflow.js`**
- Parse `args.tracelens` into `TRACELENS_INPUTS` and thread it into the
  **baseline Profile** and **first Strategize** role inputs only (post-config
  reprofile/re-strategize are intentionally excluded — the baseline prior is
  stale once config changes). Empty when absent ⇒ byte-identical.

**`roles/profiler.md`**
- New "Step 0 — TraceLens fast-path" (PHASE=baseline / ROUND 0 only):
  - If `analysis_md` exists → **skip internal trace collection** and build the
    canonical `profile_topN.json` from `kernel_candidates.json`/`tracelens_report.json`.
  - If `trace_file` also exists → run an **additional `parse_profile.py` pass**
    on the top-level **rank0** serving trace (ignoring `capture_traces/` graph
    warmups) and reconcile: prefer the parser's per-launch shapes (the mandatory
    shape double-check, since analysis.md shapes are unreliable).
  - Missing/empty, or any reprofile → unchanged collection.

**`roles/system_architect.md`**
- New advisory step 1b: read `kernel_candidates.json` as a routing prior;
  cross-check measured heads and enrich `head_candidates`/`kernel_candidates`
  with `source_hint` / `launcher_hint` / `bound_type`. Never overrides the
  on-box measured `pct_gpu_time`.

**`roles/kernel_extractor.md`**
- Use `KERNEL.source_hint`/`launcher_hint` as the first place to locate the
  source/seam (still confirmed by import+grep, never trusted blindly).
- Mandatory TraceLens shape double-check against a live capture before freezing
  the unittest.

**`interface/run_e2e.md`**
- Document the new `args.tracelens` mapping row + a "TraceLens prior
  auto-discovery" section (discovery globs, resolution order, and how the
  workflow consumes it). `handoff.json` contract is unchanged (`schema_version`
  stays `1`); `tracelens` is derived from the existing `exp_root`.

## Compatibility

- No new required handoff field; `tracelens` is derived from `exp_root`.
- Absent artifacts ⇒ key omitted ⇒ unchanged behavior. Reprofile rounds ignore
  the (stale) baseline prior.
- New candidate fields ride on `additionalProperties: true` schemas, so existing
  StructuredOutput validation is unaffected.

## Validation

- `develop_geak/tracelens_integration_check.py`: 11/11 local cases inject the
  correct non-null `args.tracelens`; bogus `exp_root` omits the key (null-safe)
  — ALL PASS.
- `develop_geak/dump_full_prompt.py`: full `build_prompt()` output regenerated
  for all 11 cases.
- Ran `parse_profile.py` on a real roofline trace: it recovers the **real**
  kernel symbol `fused_moe_kernel_gptq_awq` (71.5% GPU, editable) + concrete
  per-launch shapes, vs TraceLens's pseudo-op `pseudo_op::moe_gptq_awq_up` +
  launcher path — confirming the value of the extra analysis + shape double-check.
- `python -c ast.parse` clean on `run_e2e.py`; rank0 trace-selection verified on
  real data.

## Test plan

- [ ] End-to-end dry run: `python interface/run_e2e.py <handoff.json> <result.json> --dry-run`
      and confirm `args.tracelens` appears when artifacts exist.
- [ ] One full e2e run on a case WITH tracelens (skip-collection path) and one
      WITHOUT (legacy path) to confirm parity.

### DIFF
--- a/e2e_workflow/e2e_workflow.js
+++ b/e2e_workflow/e2e_workflow.js
@@ -32,6 +32,21 @@ const KERNEL_WF_SCRIPT = `${KERNEL_WF_DIR}/kernel_workflow.js`;
 // EXP_ROOT = where timestamped run dirs go. Default: sibling "exp/" next to this workflow dir.
 const EXP_ROOT = String(A.exp_root || (WORKFLOW_DIR.replace(/\/[^/]*$/, '') + '/exp')).replace(/\/+$/, '');
 
+// ---- Upstream TraceLens / kernel-agent prior (OPTIONAL; forwarded by run_e2e.py as args.tracelens) ----
+// run_e2e.py resolves these paths beside the perfskills handoff and forwards ONLY the non-null ones.
+// They are a PRIOR for the Profile/Strategize/Extract phases: if analysis_md exists the Profiler skips
+// its own trace collection and builds the Top-N from TraceLens (and runs an EXTRA parse_profile pass on
+// trace_file when present); the Architect uses kernel_candidates as a routing prior. ENTIRELY ADDITIVE:
+// when args.tracelens is absent every TRACELENS_* input is '' and the run is byte-identical.
+const TL = (A.tracelens && typeof A.tracelens === 'object') ? A.tracelens : {};
+const TRACELENS_INPUTS = {
+  TRACELENS_ANALYSIS_MD: String(TL.analysis_md || ''),
+  TRACELENS_KERNEL_CANDIDATES_JSON: String(TL.kernel_candidates_json || ''),
+  TRACELENS_REPORT_JSON: String(TL.tracelens_report_json || ''),
+  TRACELENS_TRACE_FILE: String(TL.trace_file || ''),
+};
+if (TL && Object.keys(TL).length) log(`TraceLens prior present: ${Object.keys(TL).filter(k => TL[k]).join(', ') || '(none non-null)'}.`);
+
 // ---- single-kernel pass-through: if kernel_path (and no model_path), just run the kernel layer ----
 const KERNEL_PATH = A.kernel_path || '';
 const MODEL_PATH = A.model_path || '';
@@ -677,6 +692,7 @@ if (want('setup')) {
     roleAgent('profiler', 'baseline', 'Capture a warm trace and emit the standardized Top-N.', {
       EVAL_DIR, MODEL_PATH, GPU_ID: GPU_LIST[0], WORKLOAD, ROUND: 0,
       OVERLAY_PYTHONPATH: '', EXTRA_SERVER_ARGS: curFlags, EXTRA_ENV: curEnv, SKILL_DIR: WORKFLOW_DIR,
+      ...TRACELENS_INPUTS,
     }),
     { phase: 'Profile', label: 'profiler:baseline', schema: PROFILE_SCHEMA });
   log(`Baseline profiled. ${profile ? (profile.top_kernels || []).length : 0} top kernels.`);
@@ -686,6 +702,7 @@ if (want('setup')) {
     roleAgent('system_architect', 'strategize', 'Route the Top-N into config/kernel/host tracks by Amdahl.', {
       EVAL_DIR, PROFILE_TOPN: profile ? profile.profile_topN_json : '', BASELINE_THROUGHPUT: BASELINE_TPUT,
       WORKLOAD, BUDGET, HEAD_THRESHOLD_PCT, CONFIG_TUNE_ENABLED, SKILL_DIR: WORKFLOW_DIR,
+      ...TRACELENS_INPUTS,
     }),
     { phase: 'Strategize', label: 'architect:strategize', schema: STRATEGY_SCHEMA });
   kernelQueue = (strategy && strategy.kernel_candidates) ? strategy.kernel_candidates.slice() : [];
--- a/e2e_workflow/roles/kernel_extractor.md
+++ b/e2e_workflow/roles/kernel_extractor.md
@@ -26,9 +26,13 @@ You are invoked once per kernel candidate. Read first:
 
 Inputs: `EVAL_DIR`, `MODEL_PATH`, `GPU_ID`, `WORKLOAD`, `KERNEL` (the Architect's candidate:
 short_name, classification, extract_hint = the `module:attr` callable to hook, candidate_backends,
-regime), `CURRENT_FLAGS`/`CURRENT_ENV`, `SKILL_DIR`.
+regime, and — when an upstream TraceLens prior was available — OPTIONAL `source_hint` (resolved source
+file), `launcher_hint` (launcher seam), `bound_type`), `CURRENT_FLAGS`/`CURRENT_ENV`, `SKILL_DIR`.
 
-1. **Locate the source.** Resolve the kernel's defining file in the installed package
+1. **Locate the source.** **If `KERNEL.source_hint`/`KERNEL.launcher_hint` is provided (TraceLens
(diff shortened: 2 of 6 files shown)


iraj465:
## Critical review — code-comparison vs the kernel-workflow path (#306), with data

Thanks @chao-xu-spec — wiring the TraceLens prior into the e2e workflow is a real improvement and the change is genuinely additive/null-safe. A few code-grounded observations on robustness, and where this sits relative to the from-analysis kernel-workflow (#306), so the team can pick the integration path deliberately.

### 1. Apply / E2E deploy capability — the biggest gap

The e2e_workflow's deploy mechanism is **Python-only**. `e2e_workflow/scripts/overlay_setup.py` is a `sitecustomize.py` on `OVERLAY_PYTHONPATH` that (a) adds a module or (b) **rebinds one Python attribute (monkeypatch)**. Grepping that file: **zero** occurrences of `.cu/.cuh`, `AITER_REBUILD`, `hipcc`/`nvcc`, `cpp_itfs`, `jit/build`, `.so`, or recompile. So:

- A **Python-level** kernel win (Triton, an `opt_flags.py` heuristic, a monkeypatchable callable) → the overlay can deploy it.
- A **device `.cu`/CK** kernel win (e.g. aiter paged-attention / quant `.cu`, the fused `module_aiter_core*.so`) → the overlay **cannot** deploy it into the live server; there's no recompile path. `kernel_extractor.md` even gates on `editable=false` / `no_rebind_seam` and depends on a Python `target_callable` rebind seam (`module:attr`) for the Integrator to wire the result back.

For comparison, HL's `apply_and_bench` (the path #306 hands off to) routes through `apply_kernel_patch`, which **does** handle the aiter/CK `.cu` case: removes the prebuilt fused `.so`, sets `AITER_REBUILD=1`, invalidates `jit/build` + `cpp_itfs` caches, and verifies a fresh rebuild — plus warmup-discard + fixed-seed + N-rep + IQR-significance + TPOT + engagement-proof + process-scoped server teardown + byte-exact reconstruct/manifest-revert. (This is exactly the gap filed as Hyperloom #717: "v4 overlay can't apply an aiter .cu".)

**Implication:** for kernels that are device `.cu`/CK (a large share of GPU time on aiter/CK serving paths), the e2e_workflow overlay can measure a config change but cannot deploy a `.cu` kernel rewrite end-to-end. The kernel-workflow path returns a patch that HL's `.cu`-capable apply then deploys.

### 2. The TraceLens-prior change itself is good — keep this part

The profiler "Step 0 fast-path" + the **"analysis.md shapes are unreliable → re-verify against a live `parse_profile.py`/capture pass"** discipline is the right instinct, and it matches a bug we hit independently (v4 decode-ctx guessed 18401 vs real 1024; attention shapes GEMM-flattened). This shape-reconciliation rule is **layer-agnostic and worth porting** into the kernel-workflow handoff regardless of which orchestrator wins — see below.

### 3. Validation / data

- **#306 has workload-level results** posted (two workloads): Llama-3.1-8B (attention 1.0096×, quant 1.0728× director-verified; combined E2E +0.47%) and gpt-oss-120b (MoE 1.1536×, all_reduce 1.0068×; both validation_status=accepted) — each measured through HL+TraceLens + the gate-less apply_and_bench (engagement-proof=true).
- **This PR (#312) currently has no end-to-end results** attached (no perf numbers, no reviews). Before adopting the e2e_workflow as the integration driver, it would need a comparable run showing it can take a real workload trace → optimize → **deploy** → measure E2E, including at least one **device `.cu`** kernel (the case the overlay can't currently handle).

### Which path, and requested changes if e2e_workflow is chosen

- If the goal is the per-kernel boundary (orchestrator owns trace/profile/serve/explore/sweep, GEAK optimizes a dispatched kernel and returns a patch), **#306 is the tested fit** and HL's apply is strictly a superset on deploy.
- If the team wants the **e2e_workflow** to be the driver, the robustness asks before it's integration-ready:
  1. **Device-kernel deploy:** add a `.cu`/CK apply path (remove prebuilt `.so` + `AITER_REBUILD=1` + jit/cpp_itfs cache invalidation + rebuild verify) — or call HL's `apply_kernel_patch` — so a `.cu` win can actually reach the live server. Today's monkeypatch overlay can't.
  2. **Engagement proof:** assert the patched kernel is on the live serving path (rebuild markers / load banner), so a ~0% delta is distinguishable from "never applied".
  3. **Measurement hardening:** warmup-discard (cold first rep), fixed dataset seed across arms, ≥5 reps + non-overlap/IQR significance, TPOT alongside throughput. (The Integrator's interleaved-A/B + parity is good for config sweeps; the kernel deploy underneath is the weak link.)
  4. **Attach end-to-end results** on ≥1 workload incl. a device `.cu` kernel.

### Reusing #312's good parts (whichever path)

- Port the **shape-reconciliation rule** ("don't trust analysis.md shapes; verify against a live capture") into the dispatch-prompt / Seed handoff used by #306.
- Keep the **TraceLens-fast-path profiling-skip** for any standalone GEAK-e2e usage (no orchestrator) — it's a real win there.

Net: #312 is solid for the profiling-prior wiring and for standalone GEAK-e2e, but its apply layer is currently Python-monkeypatch-only and can't deploy device `.cu` kernels, which is the capability HL's apply path provides and #306's results exercise. Happy to help wire #312's shape-check into the #306 handoff or to pair on the `.cu` deploy path if the e2e route is preferred.

iraj465:
Cross-link: a code-level robustness comparison vs the per-kernel from-analysis path is on #306 (https://github.com/AMD-AGI/GEAK/pull/306). Short version: this PR's TraceLens-prior + shape-reconciliation discipline is valuable and worth keeping (and porting to the #306 handoff); the integration blocker for an e2e-driver role is that `overlay_setup.py` is monkeypatch-only and can't deploy a device `.cu`/CK kernel (Hyperloom #717), which HL's `apply_kernel_patch` handles. The gap-closure checklist is in my review above.

zihaoanllm:
Regarding the recompile issue, I will fix it in a separate PR. Thanks.

PR #325, opened 2026-06-30
fix(e2e): remove forced high-priority-GEMM head reordering (HEAD_PRIORITY)

zihaoanllm:
## What
                                                                                                                                                                      
  Removes the `HEAD_PRIORITY` mechanism introduced in #304, which force-reordered the
  head-kernel queue to optimize GEMM first.                                                                                                                           
                                                                                                                                                                      
  In FAST_MODE, when `head_priority` was set (e.g. `head_priority="gemm"`), matching head
  ops were pulled to the **front** of the head queue *before* the budget slice, so the
  dense/scaled-GEMM head was guaranteed to be selected and integrated/stacked first —
  overriding the Architect's Amdahl ranking.                                        
                                                                                    
  This PR drops that logic so head ops are processed strictly in the order the System
  Architect ranks them (by `pct_gpu_time × achievable_speedup`), with no hard-coded 
  GEMM-first bias.                        
                                                                                    
  ## Changes
                                                                                    
  `e2e_workflow/e2e_workflow.js` only (+2 / −18):                                   
  - Remove the `HEAD_PRIORITY` constant + its comment block.
  - Remove the FAST_MODE head-queue reordering block; head selection is now simply  
    `headQueue.slice(0, HEAD_BUDGET)` in Amdahl order.
                                          
  ## Why
                                                                                    
  Forcing GEMM to the front is a hard-coded heuristic that pre-empts the Architect's
  per-run prioritization. Different workloads/regimes don't always want GEMM optimized
  first; the queue should reflect the measured Amdahl priority, not a fixed op-kind rule.

### DIFF
--- a/e2e_workflow/e2e_workflow.js
+++ b/e2e_workflow/e2e_workflow.js
@@ -141,12 +141,6 @@ const HEAD_AUTHOR_MAX = parseInt(A.head_author_max != null ? A.head_author_max :
 // hits a harness fault / no-win / extraction failure, the orchestrator LOUDLY flags it (and still tries
 // the author route when a plan exists) instead of dropping the biggest lever on the floor. Default 30%.
 const HEAD_PROTECT_PCT = parseFloat(A.head_protect_pct != null ? A.head_protect_pct : 30);
-// FAST-MODE ONLY head priority (opt-in, default ''=OFF): a regex matched against each head candidate's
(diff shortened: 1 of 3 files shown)


zihaoanllm:
Related issue https://github.com/AMD-AGI/GEAK/issues/334

PR #386, opened 2026-07-31
Kernel workflow: Add FlyDSL dense gemm and group gemm gfx950 skills

IzacharyI:
## Summary

Add FlyDSL gfx950 expert skills for GEAK `kernel_workflow`.

This PR covers two kernel operator families:

1. Dense fp8 a8w8 blockscale GEMM
2. Grouped-MoE GEMM

These skills are consumed when `use_expert_skills=true`. They are advisory priors for `kernel_workflow` to reproduce and measure, not hard-coded replacements for correctness, parity, or A/B gates.

## Modifications

| Path | Change |
|---|---|
| `perf_knowledge/expert_skills/skills/flydsl_fp8_blockscale_gemm/skill.md` | Adds the gfx950 fp8 a8w8 blockscale GEMM kernel skill with validated `q_up_proj`, `down_proj`, and `qkv_proj` evidence. This is distinct from `perf_knowledge/expert_skills/skills/flydsl_fp8_gemm_playbook/skill.md`, which is a gfx942 e2e down-proj overlay playbook. |
| `perf_knowledge/expert_skills/skills/flydsl_decode_moe_stage1_blkmap/skill.md` | Adds the decode stage-1 grouped-MoE blkmap skill with phase/stage guardrails, descriptor timing scope, and runtime-signature checks. |
| `perf_knowledge/expert_skills/skills/flydsl_prefill_moe_stage2_fp8partial/skill.md` | Adds the prefill stage-2 grouped-MoE fp8-partial skill with phase/stage guardrails, scale-consistency checks, and relaxed-parity guidance. |
| `perf_knowledge/expert_skills/index.yaml` | Regenerates expert-skill selector metadata for the submitted validated skills. |
| `kernel_workflow/roles/{tech_lead,engineer,deep_engineer}.md` | Surfaces target backend/language authoring knowledge for cross-backend CK -> FlyDSL ports. |
| `perf_knowledge/operators/scaled_quant_gemm/backends/{ck,flydsl}.md` | Updates CK/FlyDSL scaled-quant GEMM guidance and expert-skill references for blockscale GEMM authoring. |
| `perf_knowledge/operators/dense_gemm/{overview,numerics}.md` | Updates dense GEMM overview and blockscale numerics references for CK -> FlyDSL workflow knowledge. |
| `perf_knowledge/operators/dense_gemm/backends/flydsl.md` | Updates dense GEMM FlyDSL backend guidance and expert-skill references. |
| `perf_knowledge/operators/fused_moe_grouped_gemm/backends/flydsl.md` | Updates FlyDSL grouped-MoE backend knowledge surfaced during workflow authoring. |
| `perf_knowledge/quantization/block_scaling_mxfp.md` | Updates block-scaling guidance to reference the current blockscale GEMM expert skill. |

## Skill Organization

The submitted skills are kept separate by workflow layer, GPU generation, and kernel scenario.

`flydsl_fp8_blockscale_gemm` is separate from the existing `flydsl_fp8_gemm_playbook`: the existing playbook is a gfx942 e2e down-proj overlay focused on capture-safe integration, while the new skill is a gfx950 `kernel_workflow` recipe for CK/production blockscale GEMM -> FlyDSL candidates.

The two Grouped-MoE skills are also kept separate: `flydsl_decode_moe_stage1_blkmap` targets decode / small-M stage-1 gate/up, while `flydsl_prefill_moe_stage2_fp8partial` targets prefill / large-M stage-2 down/reduce. They have different runtime signatures, validation scopes, and parity requirements.

## Workflow Impact

- Expert skills are only consulted when `use_expert_skills=true`.
- Matched skills are advisory candidates for kernel workflow reproduction.
- Final acceptance still depends on on-box A/B measurement and parity validation.
- Default behavior with expert skills disabled is unchanged.
- CK -> FlyDSL ports now surface target backend/language knowledge earlier, so workflow agents can reuse validated GEMM recipes instead of rediscovering the porting rules.

## Validation

| Operator / Scenario | Skill | `kernel_workflow` / Director-verified result | Correctness |
|---|---|---|---|
| Dense fp8 blockscale GEMM / wide-N `q_up_proj` | `flydsl_fp8_blockscale_gemm` | `~1.226x-1.237x` | Pass |
| Dense fp8 blockscale GEMM / K-light `down_proj` | `flydsl_fp8_blockscale_gemm` | `1.4541x` | Pass |
| Dense fp8 blockscale GEMM / narrow-N `qkv_proj` | `flydsl_fp8_blockscale_gemm` | `~1.05x` | Pass |
| Grouped-MoE GEMM / decode stage-1 gate/up | `flydsl_decode_moe_stage1_blkmap` | `1.0778x` | Pass |
| Grouped-MoE GEMM / prefill stage-2 down/reduce | `flydsl_prefill_moe_stage2_fp8partial` | `1.2019x` | pass |

## E2E Validation

Validation setup: `lmsysorg/sglang-rocm:v0.5.17-rocm720-mi35x`, `DeepSeek-V4-Pro`, `sglang`, ISL/OSL `51200/1024`, concurrency `64`, num prompts `128`, GPUs `0,1,2,3,4,5,6,7`, TP/DP `8/8`.

- The MoE skill stack was validated end-to-end on the above setup, achieving approximately `3.1%` TTFT improvement and `2.6%` TPOT improvement.
- CK-to-FlyDSL & FlyDSL dense GEMM path improves TTFT by `2.4%` and TPOT by `1.6%`.

## Checks

- Static validation passes for all submitted skills.
- `index.yaml` matches skill frontmatter.
- `git diff --check` passes.
- No stale expert-skill links remain under `perf_knowledge/expert_skills`.
- No local validation artifact paths remain in the submitted skill files.
- No co-author trailers are present in the submitted commits.


yueliu14 (approved):

iraj465 (commented):
Pulled the branch, ran the validator and the new test, regenerated `index.yaml`, and traced the selector path. The work looks solid. Six issues below — **one I'd hold merge for**, the rest are cheap fixes.

---

### 1. The two MoE skills may never be selected ⛔

Both declare `match.operator: grouped_gemm_moe`. Every existing MoE skill uses `fused_moe_grouped_gemm`, and `tech_lead.md:76` — which this PR edits — lists that id in its exemplars. `expert_skills.md:19` matches on equality, so on a real run I'd expect these two to be skipped.

Both ids are legal operators, which is why `--static` passes. `validate_skill.py:60-64` already accepts a list; I tested it:

```
STATIC OK: flydsl_decode_moe_stage1_blkmap (operator=['grouped_gemm_moe', 'fused_moe_grouped_gemm'])
```

Same class on the dense skill: it declares `dense_gemm`, but this PR cross-links it from `scaled_quant_gemm/backends/ck.md`. Worth checking a real `analysis.json` for the id the tech lead actually lands on — you'll know better than me.

### 2. `--emit-plan` emits `target_language=triton` for a FlyDSL skill

`from_backend`/`to_backend` are empty and `emit_plan()` falls back to Triton. Setting both to `flydsl` fixes it.

### 3. The e2e gate is unreachable

All three skills set `e2e_delta_min_pct: 1.0` and leave `measured.e2e_pct` empty. `record()` only checks `isolated` when `scope == kernel`, so that branch never runs. The e2e numbers live only in the PR description.

Please also state the repeat count. We measured ~2.3% between-device spread on the claw campaign, which straddles the 1.6% TPOT claim.

### 4. "Default behavior with expert skills disabled is unchanged" isn't quite right

The three role edits (~22 lines) aren't behind `use_expert_skills` — role content is the prompt base, so every run picks them up. `test_expert_skills_off_identical.js` can't cover this by construction. The content itself looks like an improvement; just worth saying so in the description.

### 5. The new test never runs

`test_flydsl_skill_portability.py` isn't in `ci-l0-checks.yml`, which enumerates files rather than discovering them. Adding it also needs `pyyaml` on the pytest install line. It covers only the two MoE skills.

### 6. The dense skill ships no evidence file

So its strongest number (1.4541x on `down_proj`) can't be audited. An equivalent YAML with per-run latencies would fix that and let the portability test cover all three. The frontmatter also lags the prose — `measured.isolated` omits down_proj, though the body documents it properly.

---

### What holds up

- Static validation passes on all three — ran it.
- `index.yaml` regenerates byte-identical under `scaffold.py --reindex`.
- Separation from `flydsl_fp8_gemm_playbook` is mechanical, not just argued: `scope: e2e`/`gfx942` vs `scope: kernel`/`gfx950`, disjoint on two axes.
- Every new wikilink and language dir resolves.
- **The two MoE evidence YAMLs are the best validation artifacts I've seen go into this repo.** I recomputed every figure: prefill 1.2099x / 17.35%, decode 1.1043x / 9.45%, component medians summing to the segment median to the digit. Effects are 10-20x the run-to-run spread. Shipping raw per-run latencies rather than just the ratio is exactly right — I'd like it to become the norm.
- The E8M0-vs-arbitrary-fp32 distinction is stated consistently across all four files that touch it.

### Minor

- Validation blocks are hand-written, not written by `--record`, so `status: validated` was asserted rather than computed. Pre-existing pattern, not introduced here.
- `validate_skill.py --static` isn't run in CI anywhere, so "static validation passes" is true today with nothing keeping it true.
- Decode evidence has a 0.02% median-of-sums vs sum-of-medians gap. Harmless, but the frontmatter leans on that component split for its "1.124x GEMM-only vs 1.104x with producer" argument.
- Heads-up, not a review point: `scaffold.py --reindex` drops the comment on `ensure_flydsl` and flips it to `draft`. Already on main.
- `tech_lead.md` cites `ck -> ck_tile` as a port example, but there's no `ck_tile` language dir.

---

None of this touches the substance, and the isolated MoE measurements are genuinely well done. Item 1 is the only one I'd hold for — cheap to check now, expensive to discover later as "we shipped two skills that never fired."

zihaoanllm (approved):

menghongtao (approved):

PR #388, opened 2026-08-03
kernel_workflow: schedule GPUs on demand instead of pinning engineers

iraj465:
## Problem

Engineers are bound to a GPU by index (`GPU_LIST[i % n]`) before anyone knows how long their work will take. Task cost varies ~23x within a single round, so a pinned round ends with the slowest **GPU** rather than the slowest **task** — whichever GPU draws short work sits idle for the rest of the round. On a shared box it is also fragile: `GPU_LIST[0]` has no fallback when GPU 0 already has a tenant.

## Change

`gpu_lock.sh` accepts a pool (`"0,1,2,3"`) in addition to a single id, and takes whichever GPU is both unlocked and idle at acquire time, holding it exclusively for the command. Idleness is read from the kernel driver via sysfs, not from our own locks, so a foreign tenant is visible too. (`amd-smi` is deliberately not used — it returns empty output while another process holds the GPU, i.e. it fails exactly when it is needed.)

The pool is whatever the caller passes, so this scales from 1 to 8 GPUs with no code change.

## Measured result

76 runs across 15 kernels on MI350X. Every number is the director's independent re-measurement against the frozen baseline, arms paired per kernel:

`n` = the number of **kernels** in the comparison. Each kernel contributes one paired run per arm (same kernel, same baseline, same budget, only the GPU setting differs), so n=14 means 14 kernels x 2 arms = 28 runs behind each row. `better / worse` counts those 14 kernels individually.

| | verified geomean | n (kernels) | better / worse |
|---|---|---|---|
| 4 GPUs pinned → 2 GPUs scheduled | 2.115x → 2.138x | 14 | 7 / 7 |
| 4 GPUs pinned → 1 GPU scheduled | 2.115x → 2.125x | 14 | 7 / 7 |

**The same optimization quality on a quarter of the hardware.** The differences are inside run-to-run noise, and the win/loss split is even, which is the point: freeing 2-3 GPUs costs nothing measurable.

GPU busy time over the same runs, sampled at 1 Hz across every GPU each cell held:

| arm | busy | idle |
|---|---|---|
| 4 GPUs, pinned | 4.6% | 95.4% |
| 2 GPUs, scheduled | 10.0% | 90.0% |
| 1 GPU, scheduled | 16.8% | 83.2% |

Utilization is low in all arms because these runs are dominated by agent thinking and compile time, not kernel execution — which is exactly why holding 4 GPUs buys so little.

### Recommended setting

**The default 2-3 directions per round on 1-2 GPUs.** Additional GPUs buy wall-clock, not speedup.

## Also included

- **`directions_per_round`** (default `null` = unchanged) for callers who want a fixed fan-out. It sets `DIRECTIONS_REQUIRED` for the planner *and* hard-caps the dispatch. Both halves are needed: the cap alone only truncates a short plan, and the prompt alone gets ignored — across 31 real `plan_round` returns the planner never issued more than 3 directions.
- **Colliding directions are dispatched last, not first.** 24 of those 31 plans (77%) issued directions sharing an identical `focus_files` set. Unique `(specialty, focus_files)` pairs now go first so the integrator gets stackable patches ahead of colliding ones. They are **reordered, never dropped**: on a single-file kernel every direction carries `focus_files=[kernel.py]`, the key collapses to `specialty` alone, and dropping would silently shrink the round below what the planner asked for.

## Compatibility

- `gpu_lock.sh <single_id>` keeps its exact previous behavior and code path. Verified both modes acquire and export `HIP_VISIBLE_DEVICES` correctly.
- With `directions_per_round` unset, the planner sees no new field and the round cap stays `BUDGET_REMAINING`, as before.
- `gpu_mode: 'pin'` restores index pinning, so the scheduler can be A/B'd as a single variable rather than across two scripts.
- `node --check` and `bash -n` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kernel_workflow/kernel_lane.js
+++ b/kernel_workflow/kernel_lane.js
@@ -1,7 +1,7 @@
 export const meta = {
   name: 'kernel-lane',
   description: 'SINGLE-LANGUAGE kernel optimization worker (Director/TechLead/specialist Engineers) with budget-controlled rounds, independent verification, and integration. Optimizes ONE kernel in ONE language (mode=optimize) or authors a fresh seed then optimizes it (mode=author). This is the worker invoked per lane by the kernel-workflow dispatcher (kernel_workflow.js) and by e2e_workflow; prefer calling kernel-workflow directly unless you specifically want one unchanged lane. Target: AMD Instinct MI-series GPUs (MI300X/300A/308X/325X on CDNA3 gfx942, MI350X/355X on CDNA4 gfx950 — the target card is auto-detected on-box).',
-  whenToUse: 'Internal single-language worker. Prefer the kernel-workflow dispatcher (kernel_workflow.js) as the entry point; invoke this directly only to run one unchanged lane. Pass args.kernel_path (required), args.mode, args.target_language, args.budget, args.gpu_ids, args.task.',
+  whenToUse: 'Internal single-language worker. Prefer the kernel-workflow dispatcher (kernel_workflow.js) as the entry point; invoke this directly only to run one unchanged lane. Pass args.kernel_path (required), args.mode, args.target_language, args.budget, args.gpu_ids, args.gpu_mode, args.task.',
   phases: [
     { title: 'Setup', detail: 'director builds the isolated eval dir + canonical workspace' },
     { title: 'Author', detail: 'author_engineer writes a fresh optimize-loop seed (only when mode=author); speedup denominator stays the frozen online kernel' },
@@ -79,6 +79,20 @@ const DEEP_COST = (() => {
 })();
 const GPU_IDS = String(A.gpu_ids != null ? A.gpu_ids : '0');
 const GPU_LIST = GPU_IDS.split(',').map(s => s.trim()).filter(Boolean);
+// Every GPU consumer gets the WHOLE pool rather than a pinned lane. gpu_lock.sh
+// resolves a comma spec by flocking whichever lane is free AND idle at acquire
+// time, so placement follows what work actually costs instead of an index fixed
+// before the cost is known. Measured task costs span 23x on campaign20, and a
+// pinned round ends with the slowest LANE, not the slowest task -- at 4-way that
+// idles ~43% of lane time. Pinning is also fragile on a shared box: GPU_LIST[0]
+// has no fallback when lane 0 has a foreign tenant.
+// gpu_mode='pin' restores the pre-scheduler behavior (direction i pinned to GPU_LIST[i % n]).
+// It exists ONLY so the scheduler can be A/B'd as a single-variable change against the arm it
+// replaced; leaving it out would make the "before" arm unreachable and force the comparison to be
+// made across two different scripts, where any other drift would be indistinguishable from the
+// scheduler's effect. Default 'pool'.
+const GPU_MODE = String(A.gpu_mode || 'pool') === 'pin' ? 'pin' : 'pool';
+const GPU_POOL = GPU_MODE === 'pin' ? GPU_LIST[0] : GPU_LIST.join(',');
 const TASK = A.task || '';
 const EVAL_DIR_OVERRIDE = A.eval_dir || '';
 const APPLY_TO_ORIGINAL = String(A.apply_to_original != null ? A.apply_to_original : 'false');
@@ -486,7 +500,7 @@ if (MODE === 'author') {
   const authored = await agentT(
     roleAgent('author_engineer', 'author', 'Write the simplest correct baseline in the target language.', {
       TARGET_LANGUAGE, OP_SPEC, WORKSPACE: CANONICAL, TASK_DIR: KERNEL_PATH_ORIG,
-      GPU_ID: GPU_LIST[0], SKILL_DIR: WORKFLOW_DIR, COMMANDMENT, KERNEL_KNOWLEDGE_DIR,
+      GPU_ID: GPU_POOL, SKILL_DIR: WORKFLOW_DIR, COMMANDMENT, KERNEL_KNOWLEDGE_DIR,
     }),
     { phase: 'Author', label: `author:${TARGET_LANGUAGE}`, schema: AUTHOR_SCHEMA });
   if (!authored || !authored.authored || !says(authored.correctness, 'pass')) {
@@ -527,7 +541,7 @@ const KK_REFS = (analysis && Array.isArray(analysis.kk_refs)) ? analysis.kk_refs
 phase('Benchmark');
 const bench = await agentT(
   roleAgent('benchmark_engineer', 'setup', 'Build the COMMANDMENT and record a reliable baseline.', {
-    WORKSPACE: CANONICAL, EVAL_DIR, SKILL_DIR: WORKFLOW_DIR, GPU_ID: GPU_LIST[0],
+    WORKSPACE: CANONICAL, EVAL_DIR, SKILL_DIR: WORKFLOW_DIR, GPU_ID: GPU_POOL,
     ANALYSIS: analysis,
     ...(HARNESS_ADDENDUM ? { HARNESS_ADDENDUM } : {}),
     ...(WORKLOAD_SPEC_PATH ? { WORKLOAD_SPEC_PATH } : {}),
@@ -545,7 +559,7 @@ log(`Benchmark done. ${bench.num_test_cases || BASELINE_PER_CASE.length} cases, 
 phase('Profile');
 let profileSummary = await agentT(
   roleAgent('profile_engineer', 'baseline', 'Profile the baseline and classify the bottleneck.', {
-    WORKSPACE: CANONICAL, EVAL_DIR, SKILL_DIR: WORKFLOW_DIR, GPU_ID: GPU_LIST[0], ROUND: 0,
+    WORKSPACE: CANONICAL, EVAL_DIR, SKILL_DIR: WORKFLOW_DIR, GPU_ID: GPU_POOL, ROUND: 0,
     COMMANDMENT,
     ...RESUME_INPUT,
   }),
@@ -602,7 +616,7 @@ while (dispatched < BUDGET && noImprove < MAX_NO_IMPROVE) {
     ...d,
     idx: i,
     id: d.id || `r${round}_d${i}`,
-    gpu_id: GPU_LIST[i % GPU_LIST.length],
+    gpu_id: GPU_MODE === 'pin' ? GPU_LIST[i % GPU_LIST.length] : GPU_POOL,
     out_dir: `${EVAL_DIR}/round_${round}/engineer_${i}`,
   }));
   // deep_explore is a DEDICATED-ROUND, heavyweight mandate: if the plan includes one, run ONLY it this
@@ -728,7 +742,7 @@ Return ONLY the worker_result.json structure as StructuredOutput.` +
     integrate = await agentT(
       roleAgent('integrator', 'integrate', 'Combine this round\'s verified patches into one best implementation.', {
         CANONICAL, INTEGRATE_DIR: `${EVAL_DIR}/round_${round}/integrate`,
-        GPU_ID: GPU_LIST[0], SKILL_DIR: WORKFLOW_DIR, COMMANDMENT, BASELINE_PER_CASE,
+        GPU_ID: GPU_POOL, SKILL_DIR: WORKFLOW_DIR, COMMANDMENT, BASELINE_PER_CASE,
         BEST_INDIVIDUAL: Math.max(...candidates.map(c => c.geomean)),
         PATCHES: verified.map(r => ({ id: r.d.id, specialty: r.d.specialty, title: r.d.title,
           strategy: r.eng ? r.eng.strategy : '', verified_geomean: r.ver.verified_geomean,
@@ -790,7 +804,7 @@ re-check is not required.) Return JSON {committed, current_best_diff, note}.`,
     // --- (f) Re-profile the new best ------------------------------------
     profileSummary = await agentT(
       roleAgent('profile_engineer', 'reprofile', 'Re-profile the new best and explain the bottleneck shift.', {
-        WORKSPACE: CANONICAL, EVAL_DIR, SKILL_DIR: WORKFLOW_DIR, GPU_ID: GPU_LIST[0], ROUND: round,
+        WORKSPACE: CANONICAL, EVAL_DIR, SKILL_DIR: WORKFLOW_DIR, GPU_ID: GPU_POOL, ROUND: round,
         COMMANDMENT, PREVIOUS_METRICS: profileSummary,
       }),
       { phase: 'Optimize', label: `reprofile r${round}`, schema: PROFILE_SCHEMA });
@@ -850,7 +864,7 @@ const report = await agentT(
 phase('Validate');
 const validation = await agentT(
   roleAgent('director', 'validate', 'Independently validate the final patch vs the TRUE baseline.', {
-    KERNEL_PATH_ORIG, EVAL_DIR, WORKSPACE: CANONICAL, SKILL_DIR: WORKFLOW_DIR, GPU_ID: GPU_LIST[0],
+    KERNEL_PATH_ORIG, EVAL_DIR, WORKSPACE: CANONICAL, SKILL_DIR: WORKFLOW_DIR, GPU_ID: GPU_POOL,
     APPLY_TO_ORIGINAL, COMMANDMENT,
     FINAL_PATCH: report ? report.final_patch : `${EVAL_DIR}/final_patch.diff`,
     TECH_LEAD_REPORTED_GEOMEAN: report ? report.final_speedup_geomean : cumulative,
--- a/kernel_workflow/kernel_workflow.js
+++ b/kernel_workflow/kernel_workflow.js
@@ -1,7 +1,7 @@
 export const meta = {
   name: 'kernel-workflow',
   description: 'Single ENTRY POINT for kernel optimization on AMD Instinct MI-series GPUs (CDNA gfx942/gfx950, auto-detected on-box). Dispatches on args.mode: optimize/author -> delegate one unchanged single-language lane to the kernel_lane worker (backward compatible); bakeoff -> freeze the input kernel into ONE immutable oracle + frozen baseline, discover per-language existing impls + offline-tune env backends (aiter/CK), then run one worker lane per backend language (HIP/Triton/FlyDSL/CK/...) in parallel over the GPU pool and pick the fastest verified result across ALL candidates (author/optimize lanes AND the tuned env backend) — every one scored against the SAME frozen original baseline (anti-cheating). Wraps the unchanged kernel_lane worker (one workflow() nesting level; the dispatcher is the bake-off orchestrator).',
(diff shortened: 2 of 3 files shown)


iraj465:
# Case 1 in detail: can the scheduler do the same work on fewer GPUs?

This is the full write-up for **Case 1 only** — *same engineer count, fewer GPUs*. Case 2 (more engineers on the same GPUs) is still running and will be posted separately.

---

## 1. What was actually broken

Before this PR, each engineer was assigned a GPU by its position in the list:

```js
gpu_id: GPU_LIST[i % GPU_LIST.length]   // engineer 0 -> gpu0, engineer 1 -> gpu1, ...
```

That decision is made **before any work starts** — before anyone knows how long each direction will take. But directions in the same round are wildly uneven: one engineer tries a one-line autotune change, another rewrites the whole memory layout. Measured spread within a round is up to **23x**.

### Pinned: engineers wait on a GPU they cannot use

```
        engineer A (memory,  long)  ███████████████████████████████  32 min
        engineer B (compute, short) █████                             5 min
        engineer C (host,    short) ████                              4 min
        engineer D (algo,    med)   ██████████                       10 min

  gpu0  ███████████████████████████████                              A
  gpu1  █████ . . . . . . . . . . . . . . . . . . . . . . . . . . .  B, then IDLE 27 min
  gpu2  ████ .. . . . . . . . . . . . . . . . . . . . . . . . . . .  C, then IDLE 28 min
  gpu3  ██████████ . . . . . . . . . . . . . . . . . . . . . . . . .  D, then IDLE 22 min
        |------------------ round ends when A ends ----------------|
```

The round ends with the slowest **GPU**, not the slowest **task**. gpu1-3 finish early and sit idle — they cannot help A, because A is pinned to gpu0. Three GPUs are held for the whole round and do nothing for most of it.

### Scheduled: the same work, packed onto fewer GPUs

```
  gpu0  ███████████████████████████████                              A
  gpu1  █████|████|██████████                                        B, then C, then D
        |------------------ round ends when A ends ----------------|
```

Same wall-clock, **half the hardware**. Work is handed to whichever GPU is free when the work is ready, instead of being assigned before anyone knows the cost.

There was also a second-order failure on a shared box: `GPU_LIST[0]` has no fallback. If GPU 0 already has a tenant, the pinned engineer lands on it anyway and the measurement is contaminated.

---

## 2. How the fix works, step by step

`gpu_lock.sh` now takes a **pool** (`"0,1,2,3"`) as well as a single id. Every GPU consumer gets the whole pool, and selection happens at acquire time:

```
  engineer needs a GPU
          |
          v
  +-------------------------------------------+
  | for each gpu in pool:                      |
  |                                            |
  |   1. flock -n  ..... already taken? -> next|
  |          |                                 |
  |          v  got the lock                   |
  |   2. is it IDLE?  (sysfs, kernel driver)   |
  |      busy% <= 5 AND vram <= 1024MB         |
  |          |                                 |
  |     no --+--> unlock, try next gpu         |
  |          |                                 |
  |         yes                                |
  |          v                                 |
  |   3. hold the lock for the whole command   |
  |      export HIP_VISIBLE_DEVICES=<gpu>      |
  +-------------------------------------------+
          |
     none free? -> sleep 0.2s, retry until GEAK_GPU_POOL_WAIT (default 1200s), then fail loudly
```

Three details that matter:

**Lock first, then check idle.** Checking idleness before taking the lock is a race — another engineer can grab the GPU in between. We take the lock, *then* look; if it turns out to be busy we release and move on.

**Idleness comes from the kernel driver, not from our own locks.** Reading `/sys/class/drm/renderD*/device/gpu_busy_percent`. Our `flock`s only know about *our* jobs; a colleague's container is invisible to them but will still wreck a timing. Measured on an idle MI350X: `gpu_busy_percent=0`, `mem_info_vram_used=284MB`. Under load: `100%` / `1837MB`. Thresholds of 5% and 1024MB sit clearly between the two.

**`amd-smi` is deliberately not used.** It returns *empty output* while another process holds the GPU — it fails exactly when you need it to work. sysfs always answers.

---

## 3. Results: 76 runs, 15 kernels, MI350X

Every number below is the **director's independent re-measurement** against the frozen baseline — not the optimizer's self-report. Arms are paired per kernel, and all correctness gates passed in all three arms.

**What `n` means.** `n` is the number of **kernels**, not runs. Each kernel is run once per arm on the same baseline with the same budget, changing only the GPU setting, and the three results are compared to each other. So `n=14` is 14 kernels x 3 arms = 42 runs, and `7 / 7` counts those 14 kernels one by one. Pairing per kernel is what makes the comparison meaningful — kernels differ from each other by 4x in achievable speedup, far more than the arms differ, so an unpaired average would be dominated by which kernels happened to land in which arm.

### Speedup per kernel

| kernel | 4 GPUs pinned | 2 GPUs sched | 1 GPU sched | 1-GPU vs 4-GPU |
|---|---|---|---|---|
| `fast_rms_layernorm` | 4.172x | 2.360x | 2.236x | 0.536 ▼ |
| `fused_append_shared_experts` | 3.048x | 3.230x | 4.480x | 1.470 ▲ |
| `fused_moe_mxfp4` | 2.500x | 1.919x | 2.654x | 1.062 ▲ |
| `fused_mxfp4_quant_moe_sort` | 1.026x | 1.000x | 0.992x | 0.967 ▼ |
| `fused_qk_rope_cache_mla` | 1.881x | 1.502x | 1.855x | 0.986 = |
| `fused_qkv_rope` | 1.935x | 2.358x | 2.471x | 1.277 ▲ |
| `fused_rms_fp8` | 2.694x | 2.482x | 2.777x | 1.031 ▲ |
| `gemm` | 3.526x | 2.869x | 3.197x | 0.907 ▼ |
| `gemm_a16w16_atomic` | 1.971x | 2.225x | 1.979x | 1.004 = |
| `gemm_a16wfp4` | 2.477x | 2.545x | 2.361x | 0.953 ▼ |
| `lean_atten_paged` | 1.466x | 1.656x | 1.210x | 0.825 ▼ |
| `llama_ff_triton` | 2.975x | 3.837x | 1.705x | 0.573 ▼ |
| `mla_decode` | 1.001x | 2.282x | 2.449x | 2.448 ▲ |
| `topk` | 1.494x | 1.371x | 1.512x | 1.012 = |
| **geomean (n=14)** | **2.115x** | **2.138x** | **2.125x** | **1.005** |

```
  geomean speedup, same 4 engineers throughout

  4 GPUs pinned    ████████████████████████████████████████  2.115x
  2 GPUs scheduler ████████████████████████████████████████▏ 2.138x
  1 GPU  scheduler ████████████████████████████████████████  2.125x
                   +-----------------------------------------
                   quarter of the hardware, same result
```

**Which kernels these are.** All 14 are Triton kernels from the `triton2triton` suite — single-file, hand-curated tasks. Case 1 has **not** been run on HIP kernels or on the multi-file "head" kernels extracted from live vLLM/SGLang engines; that campaign is running now but covers Case 2 only, so no HIP Case-1 numbers exist yet. See the scope note at the end.

### Is the difference real, or noise?

| comparison | ratio | 95% CI | better / worse / tie | sign test |
|---|---|---|---|---|
| 2 GPUs vs 4 GPUs | 1.011 | [0.855, 1.194] | 7 / 7 / 0 | p = 1.000 |
| 1 GPU vs 4 GPUs | 1.005 | [0.828, 1.218] | 5 / 6 / 3 | p = 1.000 |

**Both confidence intervals span 1.0 and both sign tests give p = 1.000.** There is no detectable difference in optimization quality between 4 pinned GPUs and 1 scheduled GPU.

Being explicit about what that does *and does not* mean: this is evidence that the scheduler **does not hurt**, not proof that the three setups are identical. The per-kernel spread is large (0.54 to 2.45), so the CI is wide — it cannot rule out a real effect of roughly ±18%. The claim is "you can free 3 of 4 GPUs and nothing measurably degrades," which is what the data supports.

The big per-kernel swings run in *both* directions and are mostly agent-run variance, not a GPU-count effect. `mla_decode` at 2.448 is the clearest case: the 4-GPU arm scored 1.001x — it simply failed to find anything that round, while the 1-GPU arm found a 2.449x win. `fast_rms_layernorm` (0.536) is the same phenomenon in reverse.

### GPU utilization

Sampled at 1 Hz across every GPU, with each GPU credited only to the cell exclusively holding it:

| setup | busy (>5%) | idle | GPU-hours held | **GPU-hours wasted** |
|---|---|---|---|---|
| 4 GPUs, pinned | 4.6% | 95.4% | 51.2 | **48.8** |
| 2 GPUs, scheduler | 10.0% | 90.0% | 34.7 | **31.2** |
| 1 GPU, scheduler | 16.8% | 83.2% | 17.1 | **14.2** |

```
  idle GPU-hours burned to get the same speedup

  4 GPUs pinned    ████████████████████████████████████████  48.8 h
  2 GPUs scheduler █████████████████████████▌                31.2 h   -36%
  1 GPU  scheduler ███████████▋                              14.2 h   -71%
```

**71% of the wasted GPU time is eliminated** for a result inside noise of the original.

Utilization is low in *every* arm — that is the real finding. These runs are dominated by agent thinking, compilation, and file editing, not by kernel execution. The GPU is needed in short bursts for correctness and timing. **That is precisely why holding 4 GPUs buys so little**: three of them spend ~95% of the run waiting for an LLM to finish writing code.

### The imbalance the scheduler removes

Per-GPU busy time in the pinned arm, using only samples where no other cell was running:

```
  gpu0  ██████████ 10.4%   <- engineer 0, always assigned first
  gpu1  ████        4.2%
  gpu2  ███         3.5%
  gpu3  ███         3.0%   <- 3.5x less used than gpu0
```

This is `GPU_LIST[i % n]` made visible. gpu0 takes every round's first direction; gpu3 only gets used when a round has 4+ directions, and the planner usually issues 2-3. (The scheduled arms cannot be broken down this way — they were packed 2-4 cells at a time by design, so almost no sole-occupancy samples exist for them. The pinned arm is the one where the claim needed checking, and it holds.)

---

## 4. Recommended setting

**2-3 directions per round on 1-2 GPUs** — which is the existing default direction count, with `gpu_ids` set to 1 or 2 GPUs instead of 4.

- The planner already issues 2-3 directions naturally (across 31 real `plan_round` returns it never issued more than 3).
- 1 GPU matches 4 GPUs on speedup and wastes 71% fewer GPU-hours.
- 2 GPUs is the safer pick if you want headroom for rounds that do fan out to 4 directions, at 36% fewer wasted GPU-hours.
- Extra GPUs buy **wall-clock**, not speedup: mean wall time was 64.5 min at 4 GPUs vs 73.6 min at 1 GPU (+14%). If you have GPUs idle anyway, use them; if they are contended, this is the setting to run.

No config change is needed to get the scheduler — pooling is the default. To get the old behavior back for comparison, set `gpu_mode: 'pin'`.

---

## 5. Caveats

- **The 1-GPU arm ran 4 engineers time-sharing one GPU**, not the 2-3 the recommendation names. The recommended configuration is an interpolation between the measured 1-, 2- and 4-GPU points, not a directly measured arm.
- **One box, one GPU type** (MI350X / gfx950). Cross-node behavior is untested.
- **Wide CIs.** n=14 paired kernels with high per-kernel variance supports "no measurable harm," not "provably identical."
- **Triton only, single-file only.** Every Case-1 kernel is a single-file Triton task. HIP kernels and multi-file kernels extracted from live engines (vLLM / SGLang, AITER HIP + CK templates) are **not** covered by these numbers. That matters here specifically: HIP tasks spend far more of the run compiling CK templates from source than Triton tasks do, and compile time is exactly the phase where a GPU sits idle — so if anything the scheduler should look *better* on HIP, but that is a prediction, not a measurement. A head-kernel campaign (15 kernels, 4 HIP + 11 Triton) is running now, but it was configured for Case 2 (`A0` vs `A4`) only, so it will not produce Case-1 numbers. Running Case 1 on those kernels needs the 2-GPU and 1-GPU arms added and is roughly 30 more cell-hours; happy to do it if the reviewers want it before merge."

yueliu14:
hi @iraj465 Thanks for the great work! Here are some comments: Three requests before this lands.

  1. Missing the equal-hardware control arm. Both table rows change two variables at once — GPU count (4 → 2/1) and policy (pin → schedule) — so they can't isolate the scheduler. v4 already supports 4 engineers on 2 pinned GPUs with no code change: GPU_LIST[i % GPU_LIST.length] (kernel_workflow.js:574) with gpu_mode:'pin' and gpu_ids:'0,1' gives eng0+eng2 → gpu0, eng1+eng3 → gpu1. Could you add 4 eng / 2 GPU pinned vs. 4 eng / 2 GPU scheduled, reporting wall-clock and GPU idle? 2:1 oversubscription is where the scheduler should actually earn its keep.

  2. Please add end-to-end wall-clock per arm (median across the 14 kernels + spread), including raw GPU-busy-seconds and wall-clock-seconds. Worth also reporting total time blocked in gpu_lock.sh waiting for a free GPU — that's the scheduler's cost.

  3. Needs an e2e workload check. gpu_lock.sh is shared surface (e2e_workflow/roles/kernel_extractor.md:553, roles/kernel_surgeon.md:43, scripts/adapters/vllm.sh), and e2e runs a live serving stack holding GPUs alongside isolated kernel subruns. Idleness is sampled once at acquire, so a serving GPU that's idle between benchmark passes reads as available — if the pool ever overlaps the serving set, isolated work lands on top of it and pollutes the measurement. Suggest 1~2 workloads, each as a main vs. this-PR A/B. Report e2e wall-clock and final verified e2e throughput gain for each.

sdubagun-amd (approved):

iraj465:
Thanks @yueliu14 — all three points measured. Numbers below, MI355X (gfx950), 14 head kernels, GEAK `027521f` (main) vs `50088fa` (this PR).

## 1. Equal-hardware control arm

You were right that both published rows moved GPU count and policy together. I added the arm you asked for — 4 engineers / 2 GPUs pinned — and ran it on the same pool at the same box load as the others, so it is comparable cell for cell.

12 kernels are complete in all four arms (2 of the 14 are excluded below).

| kernel | 4 GPUs pinned | 2 GPUs pinned | 2 GPUs scheduled | 1 GPU scheduled |
|---|---|---|---|---|
| `_fwd_grouped_kernel_stage1` | 1.204x | 1.066x | 1.220x | 1.176x |
| `_gemm_a16_w16_kernel` | 1.671x | 1.071x | 1.940x | 1.932x |
| `_gemm_a8w8_blockscale_kernel` | 21.679x | 1.744x | 14.944x | 19.015x |
| `_per_token_group_quant_fp8` | 1.536x | 1.426x | 1.003x | 1.013x |
| `_w8a8_triton_block_scaled_mm` | 1.644x | 8.027x | 5.411x | 1.546x |
| `chunk_scaled_dot_kkt_fwd_kernel` | 1.110x | 6.101x | 8.172x | 7.581x |
| `fused_moe_kernel` | 32.141x | 13.504x | 1.791x | 12.390x |
| `fused_moe_kernel_gptq_awq` | 3.628x | 3.866x | 6.292x | 3.923x |
| `kernel_unified_attention_2d` | 1.014x | 1.038x | 1.000x | 1.023x |
| `paged_attention_decode` (HIP) | 3.026x | 3.262x | 3.073x | 3.118x |
| `paged_attention_ragged` (HIP) | 1.135x | 1.149x | 1.169x | 1.146x |
| `wvSplitK` (HIP) | 1.000x | 1.112x | 1.003x | 1.030x |
| **geomean (n=12)** | **2.462x** | **2.370x** | **2.514x** | **2.596x** |
| HIP only (n=3) | 1.509x | 1.609x | 1.533x | 1.544x |
| Triton only (n=9) | 2.899x | 2.696x | 2.965x | 3.088x |

**The control pair, stated plainly: on kernel speedup, the scheduler makes no difference.**
2 GPUs pinned 2.370x vs 2 GPUs scheduled 2.514x is +6.1% on the geomean, but 6 kernels get
better and 6 get worse, and a paired t-test on the log ratios gives t = +0.22 on 11 degrees of
freedom (critical value 2.20); a sign test is p ≈ 1.0. That is a null result and I would not
want it read as a win. Per-kernel speedup is dominated by which optimization the engineer lands,
and the scheduler does not change that — it changes *when* an engineer gets a GPU.

The right claim for this PR is the operational one in section 2. The kernel table's job here is
to show the scheduler **does not cost anything in quality** — going from 4 GPUs pinned all the
way down to 1 GPU with the scheduler, the geomean is 2.462x → 2.596x, i.e. the same kernels get
the same optimizations on a quarter of the hardware.

## 2. Wall-clock, GPU-busy and time blocked

| arm | wall-clock median (min) | IQR | GPU-busy s | GPU wall-clock s | busy | blocked in `gpu_lock.sh` (s) |
|---|---|---|---|---|---|---|
| 4 GPUs pinned | 61.1 | 55–77 | 9477 | 192944 | 4.9% | 1340* |
| 2 GPUs pinned | 75.9 | 68–79 | 21752 | 114734 | 19.0% | 2286 |
| 2 GPUs scheduled | 63.7 | 52–73 | 6504 | 93438 | 7.0% | 710* |
| 1 GPU scheduled | 76.2 | 66–91 | 8400 | 60805 | 13.8% | 1195* |

At the 2:1 oversubscription you asked about, same hardware, policy the only difference:

- **median wall-clock per kernel 75.9 → 63.7 min** (−16%)
- **GPU-seconds held 114734 → 93438** (−19%) for the same 12 kernels
- **time blocked waiting for a free GPU 2286 s → 710 s**, i.e. the scheduler's own cost is
  *lower* than pinning's, because a pinned engineer waits on its assigned card even when a
  different card is free.

And the scheduler recovers most of the 4-GPU wall-clock (61.1 min) on half the hardware
(63.7 min), which pinning at 2 GPUs does not (75.9 min).

**Caveat, disclosed:** `gpu_lock.sh` only started recording `wait_s` on 08-05, so the three
starred rows undercount time blocked — 4 GPUs pinned has it on 1173/1880 acquisitions,
2 GPUs scheduled 1384/1877, 1 GPU scheduled 1653/1790. The 2-GPUs-pinned row is fully
instrumented. Note this asymmetry works *against* the scheduler being flattered: the
2286 s pinned figure is complete while the 710 s scheduled figure is partial, so the real gap is
if anything wider, but I would not lean on that until the instrumentation is even.

**Excluded (2 of 14):** `gemm_a8w8_blockscale` and `write_req_to_token_pool_triton` in the
2-GPUs-pinned arm — their `gpu_use.jsonl` shows 5 and 3 distinct GPUs against a 2-GPU label, so
they did not run the arm they claim and are dropped rather than reported. Two further cells
(`chunk_scaled_dot_kkt_fwd_kernel` and `paged_attention_decode`, both 2 GPUs pinned) are flagged
by the director; they are counted above under the same rule as the published table, and with all
four flagged/excluded kernels removed the 10-kernel geomeans are 2.612x / 2.088x / 2.190x /
2.290x — the same shape.

## 3. End-to-end workload check

Two workloads, each main vs this PR, run through `interface/run_e2e.py` with serving and
optimization GPUs disjoint by construction, vllm, TP=2, isl/osl 1024, conc 64, 320 prompts,
8 warmups, 3 repeats:

| workload | leg | baseline tok/s | final tok/s | speedup | parity |
|---|---|---|---|---|---|
| Qwen3-14B-FP8 | main (`027521f`) | 4669.8 | 4661.7 | 0.999x | no regression |
| Qwen3-14B-FP8 | this PR (`50088fa`) | 4660.2 | 4660.2 | 1.000x | no regression |
| gpt-oss-120b | main (`027521f`) | 1987.5 | 1987.5 | 1.000x | no regression |

**This PR does not change the end-to-end result.** Baseline and final throughput are the same on
both legs (4669.8 → 4661.7 on main, 4660.2 → 4660.2 on this PR; the 0.2% between the two legs'
baselines is run-to-run noise on the same hardware). Neither leg accepted a kernel in the e2e
window, so this is a "no change, no regression" check rather than a gain measurement — which is
what it should be: the PR is a scheduling change, and the e2e path is exactly where it must be
invisible.

The specific hazard you were pointing at is real and worth recording: `SERVING_GPU`
(`e2e_workflow.js:69`) defaults to the first TP ids **of `GPU_LIST` itself**, so with
`gpu_ids:"0,1,2,3" tp:1` the serving card sits inside the optimization pool by default, and fast
mode's `ISO = makeSem(GPU_LIST)` hands it out. Deep mode is the only path that subtracts it. I set
the split explicitly for these runs rather than relying on the default. That default is not
introduced by this PR, but the scheduler makes it easier to hit, so it is worth a follow-up.

yueliu14 (approved):

PR #395, opened 2026-08-05
Fix/editability bindability reachability

sdubagun-amd:
Fixes three downstream defects in the GEAK e2e workflow where a correctly-selected kernel was still lost before delivering an end-to-end win:

Editability — the profile classifier mislabeled editable Triton GEMMs (e.g. `_gemm_a8w8_blockscale_kernel`) as non-editable `library_gemm`, so they were dropped from optimization. Hardened the name classifier and let TraceLens's `op_to_source_patchable` signal override the name guess when present.

Bindability — an optimized kernel injected via `sitecustomize` reached only vLLM's API/engine process, not its spawned TP worker subprocesses (where the GPU work runs), yielding 0 engagement and no speedup. Added a `--worker-extension-cls` hook that re-applies the overlay inside each worker (+ per-rank engagement proof). sglang (fork-based) is unaffected.

Reachability — a bound seam the live inference path never dispatches also shows 0 engagement, but only at the final integrate A/B hours later. Now surfaced at extract time (`live_engagement_calls`) so a dead/unreached seam is flagged/skipped before the author lane.

Adds classifier regression tests; full suite (72) passes.

---

## Validation — Qwen3-14B-FP8, vLLM v1, TP=2, MI355X (gfx950)

End-to-end validation run on the exact audit case. Result: **+50.78% E2E throughput** (baseline 4182 → 6306 tok/s/GPU), crash-free. The three fixes were exercised and behaved as intended:

- **Editability:** the dominant `a8w8_blockscale` FP8 GEMM head (the audit's 36%-GPU kernel) was correctly classified and entered the optimization pipeline — not dropped as non-editable. It was ultimately optimized via a tuned-CK-GEMM kernel-switch that contributed to the E2E gain.
- **Bindability (TP=2):** the overlay finder was confirmed installed in **all 5 processes** (APIServer + 2 engine cores + 2 TP workers); pre-fix it reached only the APIServer. A candidate that rebound only in the frontend and not the TP workers was **correctly rejected as `no_engagement`** by the per-rank gate — instead of being banked as a false win.
- **Reachability:** the extract-time probe produced a live `live_engagement_calls` signal, as designed.

### Pre-fix behavior these fixes address (same model, earlier runs)

- **Editability mislabel:** the same `_gemm_a8w8_blockscale_kernel` head was classified oppositely based only on its symbol string — `library_gemm`/`editable=False` at 36.15% GPU (run `20260630T150552Z`, autotune-suffixed name) vs `triton`/`editable=True` (run `20260630T152713Z`, bare name). The higher-impact instance was the one marked non-editable.
- **0-engagement no-win:** a pre-fix vLLM TP=2 run (`20260629T164842Z`) authored a genuinely faster GEMM (`isolated_speedup 1.4335`) that yielded no E2E win — verdict `dead_end`: *"the overlay bound `aiter.gemm_a8w8_blockscale` only in the APIServer process, never in the Worker_TP0/TP1 subprocesses that actually execute the GEMM."* This is precisely the bindability failure Fix 2 closes.


yueliu14:
hi @sdubagun-amd. Thanks for the contribution. Since the PR currently provides limited context on the issue and does not include related issue logs or error messages, it is a bit difficult for me to fully understand the problem this PR is intended to address.

Could you please add more details about the issue, including under what conditions it occurs and how it can be reproduced? Also, please include e2e workload test results or Hyperloom test results after the fix to demonstrate that the PR effectively resolves the issue. Thanks!

sdubagun-amd:
Thanks, I've added the details of the earlier findings, as well as the results from Hyperloom run after the fix.

sdubagun-amd:
Closing this as the work is being continued as multiple separate PRs.

Issue #404, opened 2026-08-13
`result.json` can report a `throughput_speedup` that contradicts its own baseline/final pair


chao-xu-spec:
## Summary

`normalize_result()` in `interface/run_e2e.py` writes `throughput_speedup`,
`baseline_throughput_tok_s` and `final_throughput_tok_s` from **different measurement legs**, and
nothing asserts they agree before the file is written. Consumers that recompute
`(final - base) / base` therefore disagree with the headline number in the same file — by up to
21 points on runs we checked.

Worst case observed: reported `1.152x` where the reported pair implies `1.0066x` and the honest
same-session A/B is `1.0068x`.

## Root cause 1 — the cold-basis switch changes the numerator and the speedup, not the denominator

```python
    if geak_cold_final and cold_gate is not None and cold_gate >= 1.0:
        final_tput_out = float(geak_cold_final)
        final_basis = "cold"
        cold_geak_sp = alignment_metrics["cold_geak_speedup"]
        if cold_geak_sp is not None:
            speedup = float(cold_geak_sp)
```

`baseline_throughput_tok_s` keeps the **hot** value. Result:

```json
{ "baseline_throughput_tok_s": 2870.512,
  "final_throughput_tok_s": 2889.537,
  "final_throughput_basis": "cold",
  "throughput_speedup": 1.152 }
```

Two further problems with the cold basis itself:

* the cold legs are **unpaired** — the numerator comes from
  `<eval_dir>/validation/final/bench_summary.json` (measured hours into the session) and the
  denominator from `<eval_dir>/baseline/bench_summary.json` (session setup);
* a "cold" round run late in a session is not cold (JIT / kernel cache / page cache are warm).
  Measured penalty: **-12.62%** on the setup baseline vs **-0.02%** on the final leg.

On a run the Director closed as `validated_no_win` (empty overlay, 0-byte patch, flags identical
to the baseline seed config), the reported `1.1056x` decomposes exactly as
`0.9995 (same-session cold A/B) x 1.1062 (baseline drift)` — the entire "gain" is drift.

## Root cause 2 — the baseline ignores the Director's drift-corrected same-session base

```python
    geak_baseline = _positive_finite_float(
        wf.get("baseline_throughput_tok_s")
        or validation.get("baseline_throughput_tok_s")
        or 0.0
    )
```

This reproduces with the cold round disabled. Observed run: the Director's same-session base is
[...]

Issue #398, opened 2026-08-09
GEAK e2e: unbounded memory growth during long delegation reaches pod cgroup limit and triggers repeated OOM kills


luochen-amd:
## Summary

During a single Hyperloom task, control was delegated to GEAK e2e for 16 h 22 min. Within that one delegation window, the pod's anonymous memory repeatedly grew from a stable ~17.7 GiB baseline to **exactly the 512 GiB cgroup limit**, and the kernel fired **7 cgroup OOM kills**.

The growth does not appear to be bounded by any application-side budget - it stops only when the kernel refuses to allocate. A single excursion added roughly 480 GiB of anonymous memory in 4 minutes (~120 GiB/min sustained).

Measured peak versus the configured limit:

```
peak working_set   549754294272 bytes  = 511.9986 GiB
cgroup limit       549755813888 bytes  = 512.0000 GiB
difference                              ~1.45 MB
```

This is filed separately from #397; that issue concerns cleanup signalling, this one concerns memory growth.

## Workload context

* Model: `deepseek-ai/DeepSeek-V4-Flash-0731`
* Backend: vLLM
* GPU: 4 x MI355X
* Workload: ISL=1024, OSL=1024, TP=1
* Memory cgroup limit: 512 GiB
* GEAK ref: `v4.1.0` (cloned `--depth 1 --branch main`)
* Task wall time: 23.7 h
* GEAK delegation window: `2026-08-07 14:21:01` -> `2026-08-08 06:43:08` UTC (982 min)

Metric source: kubelet cAdvisor `container_memory_working_set_bytes` / `_rss` / `_cache`, 30 s resolution, covering the task's entire lifetime.

## Evidence 1: GEAK was the only active component in this window

The Hyperloom phase machine stayed in `KERNEL_AGENT` for the whole window, with no interleaved phase:

```
2026-08-07 14:21:01  EXPLORE -> KERNEL_AGENT
2026-08-07 14:21:01  KERNEL entry: delegating to GEAK e2e (from=EXPLORE)
                     ...... 16 h 22 min, no phase transition ......
2026-08-08 06:43:08  KERNEL_AGENT -> SWEEP
```

The parent orchestrator was effectively silent for the entire delegation, i.e. it was blocked awaiting GEAK and executed nothing of its own:

```
phase                 duration    log lines    density
PRELUDE               174 min     1172         6.7   lines/min
GEAK delegation       982 min        3         0.003 lines/min
```

[...]

PR #406, opened 2026-08-17
fix(e2e): trustworthy result.json measurement basis + Magpie launch-recipe alignment

chao-xu-spec:
## Summary

Make GEAK's Hyperloom/Magpie interface publish a speedup measured on the **same
serving stack** and the **same session basis** as the orchestrator's reference,
and recover honestly when Validate never finishes.
Refs: #404 
---

### 1. Speedup could report drift / thermal state as optimization

**Fix:** `_same_session_baseline()` prefers Validate `validation/base` → Director
`base_block` → `drift_corrected_baseline_tok_s`. Setup stays audit-only
(`setup_baseline_tok_s`, `baseline_drift_pct`). Speedup is rebuilt from the
published pair when inconsistent (`final_over_baseline`). Hot median is always
the promoted basis; cold is diagnostic-only (`BENCH_COLD_FINAL` default **0**).

When Validate is missing: do **not** invent `validation/base` — fall back to
Setup, or keep an intermediate A/B's own paired legs
(`recovered_intermediate_ab` / `no_gain_synthesis`), with explicit
`result_source` / `baseline_basis_source`.

### 2. "Same config" ≠ "same stack"

Identical flags/env can still serve a different engine if GEAK launches via its
native adapter instead of the orchestrator's Magpie script (kernel preset,
`--trust-remote-code`, gpu-mem-util).

**Fix:** `apply_bench_launcher()` + `adapters/launchers/magpie.sh` run the
orchestrator's script with overlay `PYTHONPATH` prepended. Script resolution is
**not hard-coded** to `vllm_mi355x.sh`:

`handoff.launch_server_script` → `$MAGPIE_LAUNCH_SCRIPT` /
`$MAGPIE_<BACKEND>_SCRIPT` → `launch_recipe.benchmark_script` (+ `rglob`
subdirs). Works for `vllm_mi300x.sh` / `vllm_mi355x.sh` / `sglang_mi355x.sh`.

Degrades to `native` if the script or sibling `benchmark_lib.sh` is missing.
`serving_stack` fingerprints launcher + per-leg `aiter_mentions` /
`kernel_picks`; unaligned recipes surface as `warning_recipe_unaligned`.
`MAX_MODEL_LEN` forwarded on magpie only; dual logs (`server.log` vs
`magpie_launch.log`).

### 3. Recipe `envs:` were not replayed

Same script still diverges if `PATH`/venv differs. Parse recipe `envs:` into
`RECIPE_ENV_FILE` (NUL-delimited), replay as base layer, then GEAK-owned
run-scoped names. `GEAK_STRICT_RECIPE_ENV=1` fails closed.

### 4. GPU mis-index when `N ≠ 0`

Setting both `ROCR_VISIBLE_DEVICES=N` and `HIP_VISIBLE_DEVICES=N` double-indexes
after ROCR remaps to `0..N-1`. Magpie path now sets **ROCR only** and clears
HIP/CUDA.

### 5. Magpie-path profiler host memory unbounded (#398)

Native path already passes `max_iterations`. Magpie enables torch profiler
(with memory/flops) but set no step bound. When `PROFILE=1` + vLLM, probe
`ProfilerConfig` under the recipe env and append
`--profiler-config.max_iterations` / `delay_iterations` into `EXTRA_VLLM_ARGS`
only for declared fields.

### 6. Disk recovery when Validate is missing / crashed

- Missing final + accepted intermediate A/B → `_recover_best_intermediate_win()`
  (rank by each candidate's `e2e_delta_pct`; `accepted` ≻ `stack`; Amdahl /
  parity gates; credit every distinct kernel). Never re-base recovered wins onto
  `validation/base`.
- Nothing accepted → honest `no_gain` (`final == baseline`), not a parse error.
- Measured Validate `≤1.0×` is **never** overridden — only a **missing** final
  falls back.
- `final_patch` / `final_overlay` are `""` unless they contain real hunks /
  loadable code.

---

### Unit tests

```text
pytest interface/test_run_e2e_{measurement_basis,recovery,dispatch,alignment}.py
→ 231 passed, 1 skipped
```

### Magpie vLLM profiler path

On ROCm gfx950 with Qwen3-0.6B / `BENCH_LAUNCHER=magpie`:

- Preflight: `PROFILE=1` injects `max_iterations`; `PROFILE=0` clean
- E2E: traces produced; cmdline includes Magpie `--profiler-config.*` + bound;
  `server.log` shows `Max profiling iterations reached`
- **PASS**

### Full Hyperloom → GEAK workflows (tip `bdc68856`)

| Run | Stack | Result | Evidence |
|---|---|---|---|
| `qwen3_14_fp8_20260816` | sglang + `sglang_mi355x.sh` | `disk_intermediate_win`  | Magpie sglang recipe/env/ROCR; cold off; `aligned` (+0.19%); accepted intermediate salvage (max-δ ranking, 2 kernels credited) |
| `Qwen3.5_122B_FP8_20260816` | vllm + `vllm_mi355x.sh` | `disk_stack_provisional`  | Magpie vllm recipe/env/ROCR; AITER fingerprint; stack gate kept provisional |
| `Qwen3.5_122B_FP8_20260813` | vllm | `disk_no_gain_synthesis`  | Honest no-gain when nothing accepted; recipe alignment |

Both 20260816 runs: `recipe_env.nul` present, `MAX_MODEL_LEN=6144` on cmdline,
`ROCR_VISIBLE_DEVICES=0`, `cold_runs=0`, `final_throughput_basis=hot`,
speedup self-consistent (`final/baseline`).

---

## Test plan

- [x] Unit tests above → 231 passed, 1 skipped
- [x] Magpie profiler preflight + e2e (`PROFILE=1` bound, `PROFILE=0` clean, trace + self-stop)
- [x] Full workflow sglang (14B): intermediate `accepted` recovery + recipe alignment
- [x] Full workflow vllm (122B): `stack_provisional` recovery + recipe alignment


zihaoanllm (approved):

yueliu14 (approved):

iraj465 (commented):
Spent a good while on this one. Pulled the branch, ran the tests both ways, and checked the findings against the real claw campaign artifacts (92 GEAK handoffs, 335 server logs) rather than just reading the diff. Overall I think it's a good change and the measurement-basis fixes are the right call. The tests are real too, which I checked rather than assumed: 233 pass on the branch, and if you swap main's `run_e2e.py` back underneath them 59 fail. So they're locking behaviour, not just decorating it.

A few things below, roughly in the order I'd care about them.

## The recipe replay skips EXTRA_<BACKEND>_ARGS, but reports that it didn't

This is the one I'd want sorted before merge.

To be clear up front: it's not a regression. Main never replayed recipe env at all, so these flags weren't applied there either. The problem is that this PR adds the replay, prints `replaying N recorded env var(s)`, sets `RECIPE_ENV_REPLAYED`, and then quietly undoes it for the one variable carrying the flags that matter most.

The rule is already written down in this PR, at `run_e2e.py:495-499`:

> The launcher passes both explicitly and so overrides them by ordering anyway; naming them here is what keeps RECIPE_ENV_REPLAYED honest instead of advertising a variable a later layer silently displaces.

That's exactly why `MODEL` and `TP` are in `_RECIPE_ENV_GEAK_OWNED`. `EXTRA_VLLM_ARGS` and `EXTRA_SGLANG_ARGS` aren't, but they get displaced the same way:

```
magpie.sh:98    local _extra_args="${EXTRA_SERVER_ARGS:-}"   # comes from curFlags, not the recipe
magpie.sh:179     ${_recipe_env[@]+"${_recipe_env[@]}"}      # replays EXTRA_<BE>_ARGS from the recipe
magpie.sh:192     "${_args_var}=${_extra_args}"              # later in the same env call, so it wins
```

I checked each step instead of reasoning about it:

* `env FOO=first FOO=second` gives you `second`, and a blank last assignment gives you empty.
* The variable really does survive into the replay set. I ran this PR's own `_recipe_env_block()` plus the `_RECIPE_ENV_GEAK_OWNED` filter against a live recipe (DeepSeek-V4-Flash/20260731T041213Z) and `EXTRA_VLLM_ARGS = --kv-cache-dtype fp8` comes out the other side and into `recipe_env.nul`.
* Magpie doesn't merge it with anything. Both scripts just append whatever's in the env to the server command: `vllm_mi355x.sh:95` and `sglang_mi355x.sh:87`.

For scale: 64 of the 92 recipes set this variable (16 vLLM, 48 SGLang). `_fold_serving_fidelity_flags()` does rebuild `--context-length` from the handoff's `max_model_len` in 40 of 41 cases, so that one's covered. Nothing else is, and every one of the 64 loses at least one flag:

| count | flag | count | flag |
|---:|---|---:|---|
| 34 | `--watchdog-timeout` | 4 | `--cuda-graph-max-bs` |
| 11 | `--kv-cache-dtype` | 3 | `--page-size` |
| 10 | `--block-size` | 1 | `--context-length` |
| 9 | `--moe-runner-backend` | 1 | `--mem-fraction-static` |
| 7 | `--trust-remote-code` | 1 | `--tensor-parallel-size` |
| 7 | `--attention-backend` | 1 | `--max-running-requests` |
| 6 | `--quantization` | 1 | `--enable-expert-parallel` |
| 4 | `--dtype` | 1 | `--moe-backend` |

`--kv-cache-dtype`, `--moe-runner-backend`, `--attention-backend`, `--quantization`, `--dtype`, `--block-size`, `--page-size` are the flags that decide which kernel the server actually dispatches. Which is to say they're the exact thing the new `serving_stack` fingerprint exists to catch. Right now the fingerprint would be describing a stack the launcher just changed out from under it.

There's a second-order one on the SGLang side too. `sglang_mi355x.sh:70-76` reads `$EXTRA_SGLANG_ARGS` to decide whether to add its own `--mem-fraction-static=0.8 --disable-radix-cache`. Blank the variable and Magpie puts its defaults back, so this can add config as well as drop it. One recipe in the campaign hits that.

Fix is small, something like:

```bash
local _recipe_extra=""    # pull EXTRA_${backend_uc}_ARGS back out of _recipe_env
local _extra_args="${_recipe_extra}${_recipe_extra:+ }${EXTRA_SERVER_ARGS:-}"
```

Or if GEAK is meant to own the variable outright, then just add `EXTRA_VLLM_ARGS`/`EXTRA_SGLANG_ARGS` to `_RECIPE_ENV_GEAK_OWNED` so at least the reported field stops claiming otherwise, per your own comment above.

One thing I couldn't close out: `MAGPIE_TRUST_REMOTE_CODE` and `BENCH_TRUST_REMOTE_CODE` *are* in the replay set, so those 7 `--trust-remote-code` losses might already be handled that way. Worth a look, I didn't get to it.

Same class but much milder: `MAX_MODEL_LEN` is also in the replay set and also gets overridden by ordering at `magpie.sh:191`, but only when GEAK has it set itself, since it's guarded by `${MAX_MODEL_LEN:+...}`.

## _serving_stack_signals will fall over on a big log

`run_e2e.py:1530` makes three full copies of the file:

```python
text = log_path.read_text(...)     # 1
for line in text.splitlines():     # 2, whole list materialised
"aiter_mentions": text.lower().count("aiter")   # 3
```

The `break` at `_STACK_SIGNAL_MAX_PICKS` caps the picks but not the memory. I measured 8x file size peak RSS on a 37.7 MB log (304 MB delta, 0.83s).

Server log sizes across the campaign (n=335): p50 321 KB, p90 5.0 MB, and the largest is 1174.3 MB (Kimi-K3/20260801T134446Z). That works out to somewhere around 9.5 GB RSS and 26 seconds, and this gets called twice right at the end of a multi-hour run, which is the worst possible moment to OOM.

Streaming line by line and lowercasing per line for the counter keeps the behaviour and makes the problem go away.

## There's a leftover test still asserting the old cold basis

`test_run_e2e_alignment.py:379`, `test_cold_speedup_equals_hyperloom_provisional_ratio`, still has:

```python
assert r["final_throughput_basis"] == "cold"
```

which this PR makes impossible. It only survives because its fixture (`test_results/gemma-4-26B_session/.../result.json`) isn't checked in, so it skips. It's the `1 skipped` in the run. If anyone ever stages that fixture the branch goes red. The four cold-promotion tests in `test_run_e2e_dispatch.py` were removed correctly and replaced with the inverted assertions, this one just looks like it got missed.

## Worth warning the Hyperloom side about the empty-string change

Returning `""` for hollow artefacts is right, and I went and checked every Hyperloom consumer of `final_patch`/`final_overlay`. They all use `or ""` or `.strip()`, so nothing breaks.

But `coordinator_helpers._geak_result_has_material_product` (lines 1060-1062 on the HL side) has been trivially true this whole time, because GEAK always wrote *a* path. This turns it into a real gate on day one. Across the campaign that's 80 of 295 `final_patch.diff` hollow (27%) and 43 of 53 `final/overlay` dirs marker-only (81%).

Those runs genuinely produced nothing, so the gate is doing the right thing. It's just going to look like a cliff on the dashboard, and someone will open a bug about it if we don't say anything first.

## IMPLAUSIBLE_SPEEDUP_MARGIN isn't just duplicated, it can actually diverge

```
e2e_workflow.js:194  const IMPLAUSIBLE_SPEEDUP_MARGIN = parseFloat(A.implausible_speedup_margin != null ? A.implausible_speedup_margin : 1.0);
run_e2e.py:56        IMPLAUSIBLE_SPEEDUP_MARGIN = 1.0
```

The JS one is an overridable arg, the Python one is a literal with no override. Pass `--implausible_speedup_margin` and the two layers are applying different thresholds with no way for Python to follow along. Minor, but reading it as a copy that happens to match would be wrong.

## Small one: baseline_basis_source can lie

`run_e2e.py:1812-1832`. When `_same_session_baseline()` comes back empty and the value falls through to `wf.get("baseline_throughput_tok_s")`, it still gets labelled `"setup_baseline"`, which contradicts the non-zero `baseline_drift_pct` sitting right next to it in the output.

## CI doesn't touch any of the new launch path

`ci/node/run_geak_e2e.sh` does `export BENCH_LAUNCHER="${BENCH_LAUNCHER:-native}"`, so the recipe replay and the whole magpie launcher (+130/-11, and finding #1) never get exercised by GEAK's own CI.

---

## On getting regression numbers for this

We'd like A/B numbers showing this doesn't regress on the kernels and workloads we actually run, in the same style as #388, where `gpu_mode='pin'` was deliberately kept so the scheduler change could be A/B'd as a single variable.

One wrinkle for that. This isn't purely a measurement change. It also edits three `knowledge/learned/` cards, and those get read as advisory priors by `roles/config_tuner.md:18`, `roles/system_architect.md:20` and `roles/op_benchmarker.md:30`. The `attention-backend-triton-gfx942.md` one in particular had `--attention-backend triton` carded at ★★★ +~5% e2e, and the new caution steers the agent off it for decode-bound runs. That changes what GEAK tries, and nothing in the test suite covers it.

So a `BENCH_LAUNCHER=native` before/after isn't actually a single-variable comparison as things stand. Would you be up for pulling the three knowledge cards into their own PR? That'd make this one cleanly A/B-able the way #388 was. Otherwise we'd need a third arm (cards only) to tell the two effects apart.

Either way the plan would be `ci/dispatch/run_matrix.sh verify` across the 5 models in `ci/models.tsv`, at least 3 repeats per model per arm since the between-device spread on this campaign is around 2.3% and single evals won't resolve anything. And I'd hold the `BENCH_LAUNCHER=magpie` arm until #1 is fixed, otherwise the flag displacement confounds every number that comes out of it.

iraj465:
Following up on the numbers question from my review, now that I've gone through what actually changes at runtime rather than just what gets reported.

Short version: I don't think we can merge this on code review alone, and there are three specific reasons, two of which I hadn't spotted when I wrote the review.

**The reported speedup is not comparable across the two arms.** That's by design, it's the whole point of the PR, but it means nobody should diff `throughput_speedup` between main and this branch and conclude anything. Any A/B has to compare raw `baseline_throughput_tok_s` and `final_throughput_tok_s`. Those are basis-independent.

**`bench_e2e.sh` changes what the harness executes, not just what it prints.** `BENCH_COLD_FINAL` goes from default 1 to default 0. On main the final leg ran a cold full round before the warmup+timed rounds. Now it doesn't. The warmup round is still there, so the timed rounds are still hot, and I'd expect the effect to be small. But the server enters the timed rounds having done one fewer full pass over the workload, so the cache and JIT state at the start of the timed window is genuinely different. Small is not zero, and "small" is a guess until someone measures it. It also drops one full round of wall clock per bench, which is nice but shifts the budget math.

**The three `knowledge/learned` cards change what GEAK tries.** Covered in the review, restating because it matters for arm design.

So here's what I'd like to see, cheapest first.

### Op level first, because it's nearly free

`e2e_workflow/scripts/op_bench.py` is not in this PR. I checked the file list, it's 13 files and op_bench isn't one of them. So op-level numbers should come out identical within noise on both sides, and the KEEP set on a given head op should not move.

That makes it a good cheap control. Run the same head ops on the same shapes both sides and confirm nothing moves. If something *does* move, it's the knowledge cards steering the bake-off, and we've isolated that for the price of a few op benchmarks instead of a few 16-hour e2e sessions. The `attention-backend-triton-gfx942.md` caution is the one I'd expect to show up, since it demotes a lever that was carded at ★★★.

### Then e2e through GEAK's own CI

`ci/dispatch/run_matrix.sh verify` over the 5 enrolled models:

```
MiniMaxAI-MiniMax-M2.5        (smoke)
moonshotai-Kimi-K2.6-int4
MiniMaxAI-MiniMax-M3-MXFP8
Qwen-Qwen3-14B-FP8
Qwen-Qwen3.5-27B-FP8
```

`BENCH_LAUNCHER=native` on both arms, which is what CI does anyway and what this PR now pins explicitly.

At least 3 repeats per model per arm. The between-device spread on the claw campaign is around 2.3%, and a single run per arm can't resolve anything smaller than that. If we only get one repeat, we learn nothing except "it didn't crash", which is worth knowing but isn't a regression check.

One budget note before anyone schedules this: `PERFSKILLS_E2E_TIMEOUT_S` is 57600, so 16 hours per model. 5 models × 2 arms × 3 repeats is 30 jobs. That's a lot of cluster. If that's too much, I'd rather do 2 models × 2 arms × 3 repeats than 5 × 2 × 1, since repeats are what make the comparison mean anything.

### And one real Hyperloom-driven session, which CI cannot cover

This is the gap I care most about. `run_matrix.sh` goes straight at `run_geak_e2e.sh`. It never goes through the Hyperloom coordinator. So a green CI matrix tells us nothing about:

* the empty-string return for hollow artefacts actually arriving at `coordinator_helpers._geak_result_has_material_product`, which as of this PR stops being trivially true and becomes a real gate
* the promoter consuming a `result.json` whose `final_throughput_basis` is now always `hot`
* the magpie launcher, which is the *production* path. `run_geak_e2e.sh:71` pins native deliberately, and the comment there is explicit that `run_e2e.py` flips to magpie whenever a Magpie script is derivable from the recipe. 91 of the 92 campaign handoffs have a recipe. So real Hyperloom sessions take the magpie path and CI never does.

What I'd want is one Hyperloom-driven session on an enrolled model, ideally two: one that produces a real KEEP and one that ends no-gain. That way we see the material-product gate fire both ways rather than just the happy path.

### Magpie arm last

Hold `BENCH_LAUNCHER=magpie` until the `EXTRA_<BACKEND>_ARGS` displacement from the review is fixed. Running it before then means the flag loss confounds every number that comes out of it, and we'd have to redo it anyway. Once it's fixed, one model both sides would at least put some coverage on 130 lines of launcher that currently have none.

---

We can run the CI matrix and the Hyperloom sessions on our side if that's easier. We have the claw artifacts and the enrolled models already staged. Happy to take that on, just let me know whether you want the full 5-model verify or a subset, and whether you'd prefer to split the knowledge cards out first so the arms stay single-variable.

chao-xu-spec:
@iraj465 Thank you for spending so much time reviewing the code, running the tests both ways, and validating the findings against the real claw campaign artifacts. The review identified several important gaps that directly affected serving-stack fidelity.

I addressed the code-level findings in commit `4c0a2050`.

### 1. Preserve recipe `EXTRA_<BACKEND>_ARGS`

The main blocker you identified was real: recipe `EXTRA_VLLM_ARGS` / `EXTRA_SGLANG_ARGS` reached `recipe_env.nul`, but was then overwritten by a later assignment in the same `env` command.

The Magpie launcher now:

1. extracts the current backend's `EXTRA_<BACKEND>_ARGS` from the recipe replay array;
2. removes the duplicate assignment from that array;
3. merges the layers in this order:
   - recipe arguments as the base layer;
   - GEAK accepted flags afterward for intentional overrides;
   - profiler iteration bounds last so neither prior layer can override them.

I also added regression tests that invoke the real `magpie.sh` against a fake server script and capture the final environment received by the server.

### 2. Trust, GPU-mask, and environment boundaries

The follow-up also adds the following protections:

- whole-token matching for `--trust-remote-code` and `--no-trust-remote-code`;
- enable/disable and last-token-wins semantics;
- invalid `--trust-remote-code=true/false` forms do not affect the client state;
- only the current backend's `EXTRA_<BACKEND>_ARGS` is considered;
- ROCR/HIP/CUDA masks are filtered from both recipe replay and `EXTRA_ENV`;
- non-`IDENTIFIER=VALUE` tokens are dropped before invoking `env`;
- `env --` terminates option parsing;
- multi-line `EXTRA_ENV` is processed completely rather than stopping after the first line;
- pathname expansion cannot mutate values such as `FOO=*`.

`MAX_MODEL_LEN` now also explicitly prefers the value recorded by the recipe. The handoff scalar is used only when the recipe did not record it.

### 3. Large-log memory usage

`_serving_stack_signals()` now reads the server log incrementally.

It no longer creates full-file copies through:

- `read_text()`;
- `splitlines()`;
- whole-file `lower()`.

Memory usage therefore remains bounded independently of server-log size, while preserving the existing `aiter_mentions` and bounded kernel-pick behavior.

### 4. Stale cold-basis test

The test that depended on the uncommitted `test_results/gemma-...` artifact and asserted the old cold headline has been removed.

Its replacement constructs a synthetic session under `tmp_path` and continuously verifies that:

- `final_throughput_basis == "hot"`;
- the promoted final is the hot median;
- cold measurements remain diagnostic only;
- `cold_speedup` uses a self-consistent cold pair.

The test now runs on every clean checkout and in GitHub CI instead of being hidden behind a skip.

### 5. `IMPLAUSIBLE_SPEEDUP_MARGIN`

The margin is now read and validated once on the Python side and forwarded unconditionally to the JS workflow.

The following inputs fall back to the safe default:

- `nan`;
- `inf` / `-inf`;
- negative values;
- non-numeric strings.

The live workflow and Python recovery path therefore cannot silently apply different thresholds.

### 6. `baseline_basis_source`

A fallback baseline recovered from a disk Director artifact is no longer mislabeled as `setup_baseline`. It is now labeled:

`disk_director_validation_baseline`

The Setup baseline remains available for audit, but is not misrepresented as the source of the published denominator.

### 7. Recipe YAML and replay hardening

The update also includes:

- PyYAML `BaseLoader` to preserve shell-string semantics;
- correct folded, literal, and implicit plain-scalar continuations;
- JSON colons no longer truncate continuation lines;
- malformed YAML fails closed when the parser is available;
- strict mode refuses to proceed when PyYAML is unavailable;
- invalid environment names are rejected or dropped according to strict mode;
- an earlier empty `envs: {}` no longer hides a later populated `benchmark.envs`;
- an env block containing only GEAK-owned run coordinates is valid in strict mode and does not require an empty replay file;
- `recipe_env.nul` is written through a unique temporary file, `fsync`, and atomic replacement;
- automatically minted eval directories use a microsecond timestamp plus UUID;
- missing host-local PATH components are removed before replay.

### 8. CI coverage

L0 now covers:

- repository-local recipe YAML fixtures;
- strict PyYAML/BaseLoader behavior;
- NUL replay-file assertions;
- PATH sanitization;
- GPU-mask filtering;
- recipe `EXTRA_VLLM_ARGS` preservation;
- direct `magpie.sh` fake-server tests;
- trust-state handling;
- advisory material-audit behavior when optional directories are absent.

All new tests and fixtures are repository-local or use temporary directories. They do not depend on `/shared_nfs`, the claw tree, model weights, or machine-specific paths.

### 9. Knowledge-card A/B confounder

You suggested removing the three learned-knowledge changes from this PR so the runtime A/B remains a single-variable comparison.

The affected knowledge files are now restored exactly to their state on `main`. This PR therefore no longer changes the learned priors consumed by Config Tuner, System Architect, or Op Benchmarker.

### 10. Empty material-product behavior

Returning an empty string for hollow `final_patch` / `final_overlay` artifacts is intentional.

I confirmed that the Hyperloom consumers use empty-safe checks such as `or ""` and `.strip()`, so this does not introduce a consumer compatibility break. As you noted, it does make `_geak_result_has_material_product` a real gate for the first time, so a drop in historical dashboard counts is expected. I am calling that out explicitly here so it is not mistaken for a new regression.

### Verification

Full GitHub L0 test list:

- **1146 passed**
- **1 existing environment-dependent skip**
- **8 subtests passed**
- total statement coverage: **98.38%**
- `interface/run_e2e.py`: **96.00%**

Additional runtime verification:

- Gate B: three real handoffs passed parser-to-NUL replay validation;
- Gate C: three real recipes passed through the production parser and real `magpie.sh` into a fake server;
- Gate D: a real vLLM server started successfully, passed `/health`, served a minimal completion request, and released the GPU and port cleanly.

### Remaining performance evidence

I do not want to overstate the result: the checks above establish implementation correctness and launch-path functionality, but they do not replace the requested main-versus-PR performance regression statistics.

You previously offered to run the E2E A/B using the enrolled models and claw artifacts available on your side. Now that the `EXTRA_<BACKEND>_ARGS` displacement and knowledge-card confounder have been removed, could you please help run that matrix?

My preferred minimum matrix is:

- one enrolled vLLM model;
- one enrolled SGLang model;
- `main` versus this PR;
- `BENCH_LAUNCHER=native` on both arms;
- at least three repeats per model per arm;
- compare the raw `baseline_throughput_tok_s` and `final_throughput_tok_s`, rather than directly comparing the reported speedup across branches.

If resources permit, it would also be valuable to run:

- one model with `BENCH_LAUNCHER=magpie` on both branches;
- one Hyperloom-driven KEEP session;
- one Hyperloom-driven no-gain session.

The full five-model matrix would be ideal. If cluster cost is a concern, a two-model, three-repeat subset would provide more useful evidence than one repeat across all five models.

Thank you again for the detailed review and for offering to help run these experiments.

iraj465:
Campaign evidence for problem 1, from `/shared_nfs/hyperloom-claw` (68 runs with
a `result.json`, measured 18 Aug 2026). It supports this PR and shows the
preferred baseline source is available where it is needed.

## The published triple contradicts itself on 40% of runs

62 runs publish a complete `(baseline, final, speedup)`. On **25** of them
`|speedup - final/baseline| > 0.005`:

| | n |
|---|---|
| runs with a complete published pair | 62 |
| speedup the published pair does not produce | **25** |
| `status: ok` while the published pair is a slowdown | 13 |
| `status: no_gain` while the published pair is a win | 3 |

## Your first-choice baseline exists on exactly those runs

I resolved `_same_session_baseline()`'s three sources against each run's e2e
tree:

| source | all 68 runs | the 25 bad runs |
|---|---|---|
| 1 `validation/base/bench_summary.json` | 33 | **24** |
| 2 `director_base_block` | 0 | 0 |
| 3 `director_drift_corrected` | 0 | 0 |
| none (stays on the Setup baseline) | 35 | 1 |

So the re-measured base leg is on disk for 24 of the 25 defective runs. The fix
does not depend on the fallbacks for the population that needs it.

## Worst cases

```
GLM-5.2-MXFP4/20260814T163244Z          published 0.9877  pair 1.4758  status=no_gain
Qwen3-14B-FP8/20260730T181119Z          published 1.0022  pair 0.7008  status=ok
Llama-3.1-8B-Instruct/20260805T095104Z  published 1.0456  pair 0.8358  status=ok
Qwen3-14B-FP8/20260731T161519Z          published 1.0005  pair 0.8782  status=ok
Mixtral-8x7B.../20260730T180726Z        published 1.0032  pair 1.1205  status=ok
Mixtral-8x7B.../20260814T161705Z        published 1.0138  pair 0.8984  status=ok
```

The first row is the expensive one. `0.9877` made the verdict `no_gain`, and
that verdict discarded `dsa_sparse_attn_prefill_main_kernel` — a +29.994%
same-config A/B, parity pass, `strict_non_overlap: True`, the only director
validation of that class in the campaign. Its own published pair is 1.4758.

## Two notes

1. **This is prevalence and availability, not a replay.** `normalize_result`
   takes the workflow return, and only its output is on disk, so I cannot run
   the new code over these runs and show 25 going to 0. What I can show is that
   the defect is general rather than a GLM special case, and that source 1 is
   present for 24 of 25.
2. **The cold promotion is closed by construction here, which is the right
   call.** The old cold branch replaced the ratio without replacing the
   denominator, so a cold final could be published over a hot baseline. Keeping
   cold in `alignment_metrics` only removes that shape rather than guarding it.
   I had a separate branch guarding it; it is redundant against this PR and I
   am dropping it.

`test_a_speedup_that_contradicts_its_own_pair_is_rebuilt_from_it` plus
`SPEEDUP_SELF_CONSISTENCY_TOL` is the invariant these 25 runs violate. Worth
keeping as a permanent assertion.

chao-xu-spec:
@iraj465 Following up with three measured E2E campaign cases that exercise the result invariants discussed in this PR.

These are real workflow outcomes rather than synthetic unit fixtures. They cover one honest no-gain recovery and two validated wins.

## DeepSeek-V4-Pro — honest disk-recovered no-gain

Test protocol:

- metric: aggregate output throughput
- prompts: 96
- warmups: 8
- workload: ISL 8192, OSL 1024, concurrency 32
- runtime framework: sglang
- 
Measured result:

- result source: `disk_no_gain_synthesis`
- status: `no_gain`
- baseline: **1536.006 tok/s**
- final: **1536.006 tok/s**
- speedup: **1.000000×**
- promoted basis: **hot**
- TTFT: **1519.035 ms**
- TPOT: **16.837 ms**
- accepted kernels: none
- accepted heads: none

Alignment evidence:

- current-best same-config divergence: **+1.02%**
- alignment status: **aligned**

This is the negative-path behavior the PR is intended to preserve: when no candidate survives validation, the result reports an exact self-consistent no-gain pair instead of inventing a win or returning a parse failure.

## Qwen3-14B-FP8 — validated GEMM win

Test protocol:

- metric: aggregate output throughput
- prompts: 192
- warmups: 8
- workload: ISL 8192, OSL 1024, concurrency 64
- runtime framework: vllm

Measured result:

- result source: `workflow_return`
- status: `ok`
- baseline: **2537.288 tok/s**
- final: **3382.361 tok/s**
- promoted basis: **hot**
- pair-derived speedup: **1.333062×**
- pair-derived E2E gain: **+33.3062%**
- TTFT: **175.832 ms**
- TPOT: **18.546 ms**
- output parity: **pass**

Accepted head:

- `_w8a8_triton_block_scaled_mm`
- operation: GEMM
- backend: AITER
- measured E2E delta: **+31.884%**
- isolated speedup: **1.5647×**

Alignment evidence:

- current-best same-config divergence: **−5.12%**
- alignment status: **warning**
- warning threshold: **3.0%**

## gpt-oss-120b — validated authored-attention win

Test protocol:

- metric: aggregate output throughput
- prompts: 192
- warmups: 8
- workload: ISL 8192, OSL 1024, concurrency 64
- runtime framework: sglang

Measured result:

- result source: `workflow_return`
- status: `ok`
- baseline: **3373.355 tok/s**
- final: **4706.239 tok/s**
- promoted basis: **hot**
- pair-derived speedup: **1.395121×**
- pair-derived E2E gain: **+39.5121%**
- TTFT: **2342.985 ms**
- TPOT: **11.297 ms**
- output parity: **pass**

Accepted head:

- `_fwd_kernel`
- operation: attention
- backend: HIP
- implementation: authored
- measured E2E delta: **+41.674%**
- isolated speedup: **6.5681×**

Alignment evidence:

- current-best same-config divergence: **−0.56%**
- alignment status: **aligned**


## What these results demonstrate

These three real E2E cases support the main design decisions in the PR:

1. **The same-config baseline is the meaningful comparison.**  
   The corresponding same-config divergence is **+1.02%**, **−5.12%**, and **−0.56%**.

2. **The published triple must be self-consistent.**  
   Two measured wins retain historical speedup values that their own baseline/final pair cannot produce. Rebuilding speedup from the published pair yields **1.333062×** and **1.395121×**.

3. **The correction preserves both sides of the verdict boundary.**  
   The no-gain case remains exactly **1.0×**, while the two parity-passing wins remain substantial at **+33.3062%** and **+39.5121%**.

4. **Hot remains the promoted measurement basis.**  
   All three pair-consistent outcomes use the hot final, with cold measurements remaining diagnostic only.

5. **Alignment uncertainty remains visible rather than being hidden.**  
   The −5.12% same-config divergence is retained as an explicit warning, while the +1.02% and −0.56% cases are correctly classified as aligned.

These cases are consistent with the campaign-wide finding that the defect is general rather than model-specific. They provide real E2E evidence for the baseline-selection and self-consistency invariants in this PR.

They are not a substitute for the requested repeated main-versus-PR performance matrix, but they demonstrate that the PR corrects contradictory reporting without manufacturing gains or discarding validated improvements.

iraj465 (approved):

Issue #407, opened 2026-08-17
kernel_journey writes one acceptance twice when the overlay tag and the kernel symbol differ


iraj465:
**Repo:** AMD-AGI/GEAK
**Component:** `interface/run_e2e.py` — `build_kernel_journey`
**Base commit:** `ab9dd2db`
**Severity:** wrong data in the handoff the orchestrator replays; downstream kernel counts are inflated
**Evidence snapshot:** `/shared_nfs/hyperloom-claw/`, 77 e2e cycles, 26 Jul – 17 Aug 2026

## Summary

`build_kernel_journey` builds `kernels[]` in two passes. Pass 1 writes one entry
per optimization overlay on disk. Pass 2 adds acceptances "named only in the
workflow return". Pass 2 tests that word "only" with the emitted `kernel_id`.

That test cannot work. The two passes do not name a kernel the same way:

* an overlay directory is named for its **candidate tag** — `cand_c0_triton`;
* the workflow return names the **kernel symbol** —
  `dsa_sparse_attn_prefill_main_kernel`.

The two ids never collide. So one acceptance is written twice. The first copy
carries the measured `gpu_pct` and the A/B result. The second copy carries
`gpu_pct: null` and the same `e2e_gain_pct`, and it is also marked `KEEP`.

Any consumer that counts `kernels[]` entries with `decision == "KEEP"` counts
that kernel twice.

## Where the defect is

`interface/run_e2e.py`, pass 1:

```python
short = cand.name[len("cand_"):]      # "c0_triton"  <- the candidate TAG
m = _match_profiler(short)
kid = m["kid"] if m else _canon_kid(short)
if kid in seen:
    continue
seen.add(kid)
```

`interface/run_e2e.py`, pass 2:

```python
name = str(k.get("short_name") or ...)  # "dsa_sparse_attn..."  <- the SYMBOL
m = _match_profiler(name)
kid = m["kid"] if m else _canon_kid(name)
if kid in seen:                          # never true for this pair
    continue
```

`_match_profiler` cannot rescue the pair either. It resolves a name against the
rocprofv3 table. `c0_triton` is not a profiler symbol, so pass 1 falls back to
`_canon_kid("c0_triton")` and the tag survives as the id.

## Worked example (all paths absolute, all files readable)

Session: `/shared_nfs/hyperloom-claw/GLM-5.2-MXFP4/20260814T163244Z/geak/e2e_cycle0/`

Overlays on disk:

```
[...]

PR #408, opened 2026-08-17
fix(journey): fold a return-named acceptance into its overlay entry

iraj465:
**Branch:** `fix/journey-overlay-return-alias` → `main`
**Base:** `ab9dd2db`
**Closes:** #407

## What was wrong

`build_kernel_journey` adds acceptances "named only in the workflow return"
after it has written one entry per overlay. It tests that "only" with the
emitted `kernel_id`. The two substreams do not name a kernel the same way: an
overlay directory carries the **candidate tag** (`cand_c0_triton`) and the
return carries the **kernel symbol** (`dsa_sparse_attn_prefill_main_kernel`).
The ids never collide, so one acceptance is emitted twice — once measured, once
with `gpu_pct: null` and the same `e2e_gain_pct`, both marked `KEEP`.

Full evidence, worked example, and the affected session list are in the issue.

## What this changes

`integrate_result.json` is the file that ties the two spellings together: it
records `cand_tag` **and** `short_name`. The fix reads the overlay's claim from
there and dedupes pass 2 against it.

Two small helpers, no change to any emitted field:

* `_overlay_claim(ir)` — what an **integrated** overlay says it optimized, or
  `None`. An overlay whose `gate` is not `accepted`/`stack` claims nothing, so
  a rejected or cut-off overlay still leaves a return acceptance as new
  information.
* `_claim_for(name, gain, claims)` — the claim that already covers this
  return-named acceptance. Symbol first, on the same normalization
  `_match_profiler` uses. Integrated e2e delta as the fallback, for the runs
  where the overlay recorded no usable symbol: the return copies that number
  from the overlay's own A/B rather than recomputing it, so an exact hit is the
  same measurement. The delta only folds when exactly one **unconsumed**
  overlay claims it.

A claim is consumed at most once, so one overlay can never absorb two distinct
acceptances.

This is deliberately not a name-similarity heuristic and not a rounded-gain
match. Every join is a value GEAK itself wrote to `integrate_result.json`.

## Why it is safe on real decompositions

`/shared_nfs/hyperloom-claw/Qwen3-14B-FP8/20260816T050457Z/geak/e2e_cycle0` has
one overlay (`cand_c0_aiter`, delta 7.53) and two accepted records — a
down_proj-shaped and a gate_up-shaped GEMM, deltas 6.779 and 0.705. Those are a
decomposition, not an alias. The consume-once rule keeps both entries. Verified
in the replay below and pinned by
`test_one_overlay_is_consumed_by_at_most_one_return_acceptance`.

## Verification 1 — replay over the live campaign

Rebuilt the journey for every `workflow_return.json` under
`/shared_nfs/hyperloom-claw/*/*/geak/e2e_cycle*/`, on `ab9dd2db` and on this
branch, and diffed the emitted entries.

| | before | after |
|---|---|---|
| e2e cycles replayed | 77 | 77 |
| `kernels[]` entries | 97 | **87** |
| entries removed | — | **10** |
| entries added | — | **0** |
| entries altered | — | **0** |
| `KEEP` entries | 36 | 26 |

Every removed entry carried `gpu_pct: null` and an `e2e_gain_pct` identical to a
surviving measured entry. Nothing measured was dropped. Ten sessions changed;
they are listed in the issue.

Downstream, Hyperloom's collector credited 8 of these duplicates as real
kernels, so the number on the TOP Model CI dashboard was over-counting GEAK's
kernels by 35 %.

## Verification 2 — tests

Eight new tests in the existing `interface/test_run_e2e_dispatch.py` journey
suite. Each pins one rule, and each is modelled on a real session:

| test | pins |
|---|---|
| `test_return_acceptance_already_on_disk_as_an_overlay_is_not_re_emitted` | the defect itself; the measured entry survives |
| `test_sibling_candidates_for_one_symbol_each_stay_their_own_entry` | folding is against the return, never between overlays |
| `test_symbol_less_overlay_folds_the_return_on_its_integrated_delta` | the delta fallback (Kimi-K3 case) |
| `test_an_ambiguous_delta_never_folds_two_kernels` | two overlays sharing a delta fold nothing |
| `test_one_overlay_is_consumed_by_at_most_one_return_acceptance` | the decomposition case above |
| `test_a_rejected_overlay_claims_nothing` | do-no-harm; a REVERT is not an acceptance |
| `test_an_incomplete_ab_overlay_claims_nothing` | no `integrate_result` at all |
| `test_folded_return_acceptance_is_absent_from_synthetic_discovery` | a folded entry does not return as a synthesized hot_kernel |

Full `interface/` suite, same interpreter, same machine:

```
base ab9dd2db : 168 passed, 1 skipped, 0 failed
this branch   : 176 passed, 1 skipped, 0 failed
```

## Reproduce the replay

```bash
python - <<'PY'
import json, glob, importlib.util
from pathlib import Path
spec = importlib.util.spec_from_file_location("rx", "interface/run_e2e.py")
rx = importlib.util.module_from_spec(spec); spec.loader.exec_module(rx)
n = 0
for wfp in sorted(glob.glob('/shared_nfs/hyperloom-claw/*/*/geak/e2e_cycle*/workflow_return.json')):
    d = str(Path(wfp).parent)
    j = rx.build_kernel_journey(json.load(open(wfp)), {"eval_dir": d})
    n += len(j["kernels"])
print("journey entries:", n)   # 97 on ab9dd2db, 87 on this branch
PY
```

## Not addressed here

* Pass 1 still names an entry after the candidate tag even when
  `integrate_result.short_name` holds the real profiler symbol. Changing the
  emitted `kernel_id` would move the fold key the orchestrator's assembler uses,
  so it belongs in its own change.
* Hyperloom's defensive collapse (AMD-AGI/Hyperloom `8d108b0`) stays. It has to:
  the 77 journeys already written to disk still contain the duplicates, and the
  dashboard reads those files. Once this ships, that collapse is a no-op on new
  runs.

### DIFF
--- a/interface/run_e2e.py
+++ b/interface/run_e2e.py
@@ -2215,6 +2215,53 @@ def _journey_return_entry(eval_dir: str, k: dict, idx: int, wf: dict,
     }
 
 
+def _overlay_claim(ir: Any) -> dict | None:
+    """What an INTEGRATED overlay says it optimized, or None when the overlay is
+    not integrated (rejected / A/B never completed) and so claims nothing.
+
+    ``integrate_result.json`` is the only place that records both spellings of a
+    kernel: ``cand_tag`` is the overlay directory's name and ``short_name`` is the
+    symbol the workflow return uses. A claim carries the symbol plus the
+    integrated e2e delta, and is consumed at most once (``used``) so one overlay
+    can never account for two distinct acceptances.
+    """
+    if not isinstance(ir, dict):
+        return None
+    if str(ir.get("gate") or "").lower() not in ("accepted", "stack"):
+        return None
+    gain = ir.get("e2e_delta_pct")
+    return {
+        "sym": _norm_kname(str(ir.get("short_name") or "")),
+        "gain": float(gain) if isinstance(gain, (int, float))
+                and not isinstance(gain, bool) else None,
+        "used": False,
+    }
+
+
+def _claim_for(name: str, gain: Any, claims: list[dict]) -> dict | None:
+    """The overlay claim that already covers this return-named acceptance, or None
+    when the return names a kernel no overlay on disk accounted for.
+
+    Matched on the symbol first — the same normalization the profiler match uses,
+    so a spelling difference does not split one kernel in two. Falling back to the
+    integrated e2e delta covers the runs where the overlay recorded no usable
+    symbol: the return does not recompute that number, it copies the overlay's own
+    A/B result, so an EXACT hit is the same measurement rather than a coincidence.
+    The delta is only allowed to fold when exactly one unconsumed overlay claims
+    it; an ambiguous delta never merges two kernels.
+    """
+    nk = _norm_kname(name)
+    if nk:
+        for c in claims:
+            if c["sym"] and c["sym"] == nk:
+                return c
+    if isinstance(gain, (int, float)) and not isinstance(gain, bool):
+        hits = [c for c in claims if not c["used"] and c["gain"] == float(gain)]
+        if len(hits) == 1:
+            return hits[0]
+    return None
+
+
 def build_kernel_journey(wf: dict, normalized: dict) -> dict:
     """Build the kernel_journey handoff (recorder-input shapes the orchestrator
     replays through the SBD SDK — KERNEL_JOURNEY_SCHEMA.md §2).
@@ -2273,6 +2320,14 @@ def build_kernel_journey(wf: dict, normalized: dict) -> dict:
 
     kernels: list[dict] = []
     seen: set[str] = set()  # dedup on the FINAL emitted kernel_id
+    # What each INTEGRATED overlay says it optimized (see _overlay_claim). The id
+    # dedup above cannot carry pass 2, because the two substreams name a kernel
+    # differently: an overlay dir is named for its CANDIDATE TAG (``cand_c0_triton``)
+    # and the workflow return for the KERNEL SYMBOL
+    # (``dsa_sparse_attn_prefill_main_kernel``). integrate_result.json is the file
+    # that ties the two together, so the claim is read from there, never from the
+    # directory name.
+    claims: list[dict] = []
 
     # 1) Disk truth: one entry per optimization overlay, driven by integrate_result.
     if eval_dir:
@@ -2285,25 +2340,34 @@ def build_kernel_journey(wf: dict, normalized: dict) -> dict:
                 short = cand.name[len("cand_"):]
                 if not short:
                     continue
+                ir = _read_json(cand / "integrate_result.json")
+                claim = _overlay_claim(ir)
+                if claim:
+                    claims.append(claim)
                 m = _match_profiler(short)
                 kid = m["kid"] if m else _canon_kid(short)
                 if kid in seen:
                     continue
                 seen.add(kid)
-                ir = _read_json(cand / "integrate_result.json")
                 kernels.append(_journey_overlay_entry(
                     eval_dir, short, ir, wf, geak_sha, overall_parity,
                     m["pct"] if m else None, m["name"] if m else None,
                     kernel_id_override=kid))
 
     # 2) Augment with accepted kernels named only in the workflow return (live path
-    #    / no overlay on disk), deduped against the overlay entries above (by id).
+    #    / no overlay on disk), deduped against the overlay entries above: by id,
+    #    and by the identity those overlays claimed, which is what catches the same
+    #    acceptance spelled as a candidate tag on disk and as a symbol in the return.
     accepted = list(wf.get("accepted_kernels") or []) + list(wf.get("accepted_heads") or [])
     synth_hot: list[dict] = []
     for idx, k in enumerate(accepted):
         if not isinstance(k, dict):
             continue
         name = str(k.get("short_name") or k.get("name") or k.get("op_kind") or f"kernel{idx}")
+        claimed = _claim_for(name, k.get("e2e_delta_pct"), claims)
+        if claimed is not None:
+            claimed["used"] = True
+            continue
         m = _match_profiler(name)
         kid = m["kid"] if m else _canon_kid(name)
         if kid in seen:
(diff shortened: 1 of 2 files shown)


iraj465:
## Cross-workload view: how often the twin actually fires

Per-workload breakdown of the duplicate this PR removes, measured **17 Aug 2026**
over `/shared_nfs/hyperloom-claw/` — **66 GEAK CI runs across 15 workloads**.

"Claim" below means a `KEEP` + `integrated` row in
`geak/e2e_cycle*/kernel_journey.json`, counted across **all** cycles.

| workload | runs | runs with a claim | `KEEP` rows | distinct kernels | twins |
|---|---:|---:|---:|---:|---:|
| Qwen3-0.6B | 9 | 2 | 4 | 3 | 1 |
| Qwen3-14B-FP8 | 9 | 3 | 6 | 5 | 1 |
| MiniMax-M3-MXFP4 | 7 | 0 | 0 | 0 | 0 |
| Qwen3-8B | 7 | 0 | 0 | 0 | 0 |
| gemma-4-26B-A4B-it | 7 | 3 | 9 | 6 | 3 |
| DeepSeek-V4-Pro | 5 | 0 | 0 | 0 | 0 |
| Kimi-K3 | 5 | 1 | 2 | 1 | 1 |
| Mixtral-8x7B-Instruct-v0.1 | 5 | 1 | 2 | 2 | 0 |
| Llama-3.1-8B-Instruct | 4 | 1 | 2 | 2 | 0 |
| Qwen3.5-122B-A10B-FP8 | 2 | 2 | 4 | 2 | 2 |
| Qwen3.8-2.4T-A95B-Quark-MXFP4 | 2 | 0 | 0 | 0 | 0 |
| GLM-5.2-MXFP4 | 1 | 1 | 3 | 2 | 1 |
| MiniMax-M3-MXFP8 | 1 | 1 | 4 | 3 | 1 |
| Qwen3.5-397B-A17B-MXFP4 | 1 | 0 | 0 | 0 | 0 |
| gpt-oss-120b | 1 | 0 | 0 | 0 | 0 |
| **TOTAL** | **66** | **15** | **36** | **26** | **10** |

**10 of 36 accepted-kernel rows are twins — 28%.** They appear in **8 of the 15
workloads**, in every workload that claims more than one kernel. This is not
specific to one model, one backend or one operator kind.

Two workloads are worth reading closely:

* **Qwen3.5-122B-A10B-FP8** — 4 rows, 2 real. Half the claimed kernels in this
  workload are duplicates.
* **gemma-4-26B-A4B-it** — the largest claimant, 9 rows and 3 twins. Its
  `20260815T130915Z` run shows the shape plainly:

  ```
  fused_moe_kernel_c0_triton                     gpu=35.22  gain=14.973
  fwd_grouped_kernel_stage1_c0_triton            gpu=33.59  gain=18.9148
  fwd_grouped_kernel_stage1_stage2_c0_triton     gpu=35.22  gain=14.973
  fwd_grouped_kernel_stage1_stage2_c0_triton_v2  gpu=35.22  gain=10.512
  fwd_grouped_kernel_stage1                      gpu=None   gain=18.9148   <- twin of row 2
  ```

  Row 5 is row 2 under its symbol instead of its candidate tag: no `gpu_pct`,
  identical gain. Note rows 1 and 3 also share a gain (14.973) and **both** carry
  a `gpu_pct` — two measured rows, so they are not a twin pair and both survive.

The replay result already in the PR body — 97 → 87 journey entries, 10 removed,
0 added, 0 altered — is the same 10 seen here, distributed across those 8
workloads.

### Reproduce

```bash
python3 - <<'PY'
import glob, json
from collections import defaultdict
per = defaultdict(lambda: {"runs": 0, "keeps": 0, "twins": 0})
seen_runs = set()
for kj in sorted(glob.glob('/shared_nfs/hyperloom-claw/*/*/geak/e2e_cycle*/kernel_journey.json')):
    m = kj.split('/')[3]
    rows = []
    for k in (json.load(open(kj)).get("kernels") or []):
        e = k.get("e2e") or {}
        if str(e.get("decision") or "").upper() == "KEEP" and e.get("integrated"):
            rows.append((k.get("gpu_pct"), e.get("e2e_gain_pct")))
    per[m]["keeps"] += len(rows)
    groups = defaultdict(list)
    for i, (gpu, gain) in enumerate(rows):
        groups[round(gain, 3) if isinstance(gain, (int, float)) else i].append(gpu)
    for g in groups.values():
        meas = [x for x in g if x is not None]; un = [x for x in g if x is None]
        if meas and un: per[m]["twins"] += len(un)
for m, d in sorted(per.items(), key=lambda kv: -kv[1]["keeps"]):
    if d["keeps"]: print(f"{m:32} keeps={d['keeps']:3}  twins={d['twins']:3}")
PY
```

iraj465:
## Follow-up: what the 26 claimed kernels are worth downstream

Extending the cross-workload view above, over the same **66 GEAK runs across 15
workloads** in `/shared_nfs/hyperloom-claw/`, 17 Aug 2026.

GEAK claims **26 distinct** `KEEP`+`integrated` kernels in 15 runs — 36 journey
rows before the alias twins this PR removes are collapsed.

Hyperloom credits **0 of them** with an end-to-end gain:

* `reports/kernel_optimization_summary.json` is readable in 54 of the 66
  sessions and holds 375 `by_kernel` rows — `UNATTEMPTED` 357,
  `ATTEMPTED_REJECTED` 14, `IN_FLIGHT` 4, **accepted 0**.
* In the 62 sessions with a readable `decision_trace.jsonl`, patches land as
  `change == "integrate_patch"` — `REVERT` 106, `KEEP` 29, `no_promote` 13 — and
  the row records `component: "orchestration"`, `operation_kind: "other"`, never
  the kernel it integrated. There is no `kernel` value anywhere in
  `decision.kind`.
* 15 sessions hold a GEAK `KEEP` kernel claim. 11 hold a Hyperloom
  `integrate_patch` `KEEP` with a gain. The intersection is **1**, and there the
  gain is `null`.

Four runs show GEAK and Hyperloom both positive and within 3x. In all four,
Hyperloom's whole gain is one server argument — `--max-running-requests 64`,
`--max-model-len 6144`, `--context-length 13312`, `--tp-size 1 --dp-size 2`.
None is a kernel.

This does not weaken the case for this PR. It sharpens it: the claim count is
currently inflated by 10 of 36 rows, so the first thing to get right is what
GEAK actually claims. Full measurement, including the per-workload table and the
worked GLM case, is on the Hyperloom side:
https://github.com/AMD-AGI/Hyperloom/pull/1209#issuecomment-5318026744

chao-xu-spec (approved):

sdubagun-amd (approved):
LGTM

PR #411, opened 2026-08-18
 Enforce the learned-KB contract in kernel workflow

zihaoanllm:
Builds on `feature/kernel_learned_knowledge`; 16 commits, no divergence.

## What this PR does

**Seeds the knowledge base with 135 active cards** distilled from four finished campaigns
(campaign20 at 2h / 4h / 8h per kernel, plus a 12h-per-kernel run over a second benchmark set).
Every card entered through one path: an agent reads a run's own artifacts, writes a proposal, and
`kb.py propose` lints it — a rejected proposal writes nothing. No card in this tree was hand-placed.

**Makes the read path budgeted and measurable.** `plan_round` sees at most `KB_DIR_CAP` cards and
never in round 1, so a cold first round stays profile-derived. The planner declares `learned_refs`
per direction and the script joins them against what the verifier independently measured, so
`cited_then_verified` / `became_winner` are structural facts rather than agent self-reports.

**Makes the wall-clock budget real.** `deadline_epoch` is threaded into the round loop and checked
before each round, and a TechLead stop is refused for as long as the window is materially unspent.
The prose version of the same rule had produced a 9.81x spread between the nominal window and the
realised one.

**Fixes three defects, all the same shape** — one rule with two implementations, where the wrong one
is what ran:

1. `doctor`'s cap warning grouped by `key`, which is unique per card, while eviction grouped by
   `kernel_class`. Headroom therefore read `CLASS_CAP-1` forever and `classes_at_cap` was
   structurally empty: the tree reached `attention_decode` 31/32 while doctor reported 31 free. All
   call sites now share `class_of`.
2. The citation ledger had no producer. `drain` could apply citations but nothing ever filed any,
   and the same arithmetic sat in `roles/update_experience.md` for an agent to do by hand — 292
   citations across two campaigns produced 7 recorded losses, so every card stayed at `losses: 0`
   and the ranking function could not order anything. New `kb.py cite` files the ledger from the
   lane; the prose is deleted.
3. ★★★ was unreachable. `confirms_blind` was read as "confirmed by a run with the KB off", and every
   arm runs with it on, so all 135 cards were self-confirmed-only. Blind credit is now granted when
   a card wins on a kernel other than the one it was distilled from — the distinction that actually
   guards against memorisation.

Also: confirm counters must be 0 on a new card. 129 of 135 asserted 1 at creation, which is where
112 recorded confirm-points against 49 actual wins came from.

## Verification

**Measured effect.** Director-validated speedups (`director_verified_speedup_geomean` — an
independent re-benchmark against the frozen baseline, not a TechLead self-report), geomean over the
18 kernels common to all three arms:

| arm | geomean |
|---|---|
| 2h, no KB | 2.863x |
| 2h, KB on | **3.512x**  (+23%) |

Largest gains at equal budget:

| kernel | no KB | KB |
|---|---|---|
| `fused_moe_kernel` | 16.97x | **52.71x** |
| `chunk_scaled_dot_kkt_fwd_kernel` | 2.65x | **12.34x** |
| `fused_moe_int4_w4a16` | 1.43x | **4.61x** |
| `_gemm_a16_w16_kernel` | 2.04x | **3.60x** |

On a second, disjoint benchmark set (7 kernels), a 2h run reading cards distilled from 12h runs of
those kernels reached **86% of the 12h result**; five of the seven landed at 94–100%, two of them in
a single round where the 12h runs had needed 11–13.

**Tests.** `scripts/test_kb.py` covers the index contract, the admission gates and the citation loop;
`scripts/test_lane_refs.py` checks that every ALL-CAPS identifier read in code position in
`kernel_lane.js` is declared there (written after a dispatcher-only constant referenced from the lane
killed 13 kernels at their first plan step — a parse check cannot catch a runtime reference). Both
green. Every new assertion was red before its fix, and each guard was re-checked by deleting the fix
and confirming the test goes red again.

**Tree health.** `kb.py lint --cards` audits 135 cards with 0 failing; `kb.py index --check` reports
INDEX.md in sync with the cards it indexes; the inbox is empty.


yueliu14 (approved):

Umangatamd:
## Deep review — request changes

I found several confirmed contract-breaking defects despite all CI checks passing.

### Merge blockers

1. **The KB cap and cold controls run after execution.** Round 1 receives `LEARNED_KB: on`, every returned direction executes, and only afterward are excess citations removed (`kernel_workflow/kernel_lane.js:891-899`, `1164-1185`). An over-cap direction can win and be committed, then be falsely recorded as cold/uninfluenced. This also contradicts the PR's "never in round 1" claim.

2. **The fixed-window contract still stops early.** The outer loop exits after `MAX_NO_IMPROVE` regardless of remaining wall-clock time (`kernel_workflow/kernel_lane.js:855`, `1140-1141`). Forced replanning only handles a TechLead `stop`; it does not override this deterministic early exit.

3. **Negative citation evidence is systematically omitted.** Citation filing is nested inside the successful-card gate requiring `UPDATE_EXPERIENCE_ON`, validation, and `finalPrimary > 1.0` (`kernel_workflow/kernel_lane.js:1271-1332`). Runs with speedup <=1, held-out/contended runs, and bake-off lanes therefore record no attempts or losses. Bake-off explicitly sets every lane's `update_experience: 'off'` (`kernel_workflow/kernel_workflow.js:357-374`) and never aggregates those lane citations centrally. This selects successful runs into the feedback data—the opposite of the proposed downward-pressure mechanism.

4. **Regressing and unverified directions receive incorrect citation outcomes.** `winner` means only "best candidate this round," while `improved` is a separate condition (`kernel_workflow/kernel_lane.js:1095-1105`). Citation credit nevertheless uses `became_winner`, so a 1.5x candidate can earn confirmation while the incumbent remains at 2x (`1194-1200`). A missing verifier result is also converted to `cited_then_verified: 0`, which `kb.py` records as a real loss.

5. **Flagged or unknown validation can win, be applied, and be learned.** Lane curation checks only for a validation object and speedup (`kernel_workflow/kernel_lane.js:1244-1271`); it does not require accepted validation or passing correctness. Bake-off likewise ranks candidates solely by `speedup > 1.0` (`kernel_workflow/kernel_workflow.js:392-413`). Require `validation_status === 'accepted'` and passing correctness before ranking, applying, or learning.

6. **The live writer is broken and bypasses the serialized proposal path.** `roles/update_experience.md:109-116` invokes a Python program through Node using an invalid CLI:

   ```bash
   node ${SKILL_DIR}/scripts/kb.py ... index ${LEARNED_DIR}
   ```

   It also writes cards and regenerates the full index directly from concurrent lanes, bypassing `propose`/single-writer `drain`. Even after fixing the command, stale concurrent regenerations can lose index entries. Future cards are additionally misclassified because `analysis.kernel_type` (`triton|hip|cuda|...`) is passed as the semantic `kernel_class` (`kernel_workflow/kernel_lane.js:1280-1284`).

7. **`drain` violates both the counter and merge contracts.** It merges by generated filename rather than `key`, invents one confirmation for every non-blind card, and hardcodes losses to zero (`kernel_workflow/scripts/kb.py:789-850`). In an isolated reproduction, two valid cards with the same key, zero confirmations, and two losses produced two separate files; each emerged with `confirms_cited: 1` and `losses: 0`.

8. **Cross-kernel confirmation remains nonfunctional.** All 135 committed cards lack `origin_kernels`, and the production writer does not emit it. Therefore the cross-kernel condition at `kernel_workflow/scripts/kb.py:901-904` always withholds blind credit. Ranking at `422-440` also ignores `confirms_blind`, so even synthetic blind wins do not improve standing.

### Additional defects

- Deep-explore ignores KB-off: it reads "ALL knowledge files" without receiving `LEARNED_KB` (`kernel_workflow/kernel_lane.js:964-1007`).
- `kernel_workflow_bmk.js:44-52` does not forward `use_learned_kb`, so batch controls silently use the lane's ON default.
- The generated index claims a 40-card cap but contains 135 card rows; `GEN_CAP` renders text but enforces nothing.
- `lint --cards` always returns exit code 0, even when failures exist (`kernel_workflow/scripts/kb.py:486-516`).
- The four new regression scripts are omitted from the explicitly enumerated CI test command.
- The e2e KB currently reports 23/23 cards failing lint and a stale index.

### Verification

Reviewed head `11fb331162da0c9c551bcae220f3a4eae163d24c` against `main` at `ac371a88066508fb28808d529b94267c97b2d5a7`. All nine GitHub checks pass. The four added regression scripts and kernel KB lint/index checks pass manually, but they do not cover the failures above.

zihaoanllm:
HI @Umangatamd 
Thank you for the thorough review. Several of these are in paths our own tests did not cover, and one is a regression introduced in this PR. All findings reproduced. Fixed in `ca6dc78a` and `ea32bbbe`, except #1 and #2, which are answered below rather than in code.

**1. Round 1 does read the KB — the description was inaccurate, not the code.**
There is no round-based gate anywhere in the lane, and the cap is applied after the round by stripping surplus citations, so an over-cap direction can win, be committed, and then be recorded as uninfluenced. The PR description has been corrected. What does exist is `KB_COLD_DIRECTION`: at least one direction per round must be planned with empty `learned_refs`. That is a per-round control, not a round-1 exclusion.

**2. `MAX_NO_IMPROVE` — intended behaviour, and the documentation was unclear.**
The wall-clock budget bounds when the last round may *start*; `MAX_NO_IMPROVE` bounds how many unproductive rounds to buy. They are independent stops by design. The parameter documentation has been updated to state this, and the forced re-plan only ever covered a TechLead `stop`.

**3. The ledger recorded only successful runs.**
Confirmed, and this one was introduced here: the citation filing sat inside the `finalPrimary > 1.0` curation gate, so sub-baseline, flagged and contended runs — the runs that generate losses — were filtered out before they could file anything. It has been moved outside the gate. Curating a new card still requires a verified, accepted win; the ledger of what existing cards did is owed by every run. `held_out` remains exempt, since that split is the measuring instrument.

**4. Credit was attributed to the wrong outcomes.**
`became_winner` now requires `improved` as well, so a 1.5x round winner no longer confirms a card while the committed best is at 2x. A missing verifier result is now `null` rather than `0`, and `drain` scores it as an attempt with no evidence in either direction; previously it was charged as a loss.

**5. Unverified results could win, be applied, and be learned from.**
Both the lane's curation gate and the bake-off ranking now require `validation_status === 'accepted'`. Lanes that beat the baseline but are not accepted remain in the results table and are logged, but are no longer eligible to win.

**6. The live write path could not have worked.**
`node .../kb.py ... index` invoked a Python program through Node, with an argument order the parser rejects. The role now writes a proposal and calls `propose`, leaving `drain` as the single writer, which also removes the concurrent index-regeneration race. Steps 2a and 3, the Inputs list and the ADD-only rule were inconsistent with the new step 7 and have been brought in line.

**7. `drain` did not honour its merge and counter contracts.**
Merging is now keyed on `key`, with the generated filename as a fallback for cards filed before keys were normalized. Inserts start every counter at zero and merges increment `attempts` only; the previous `confirms_cited: 1` at insert alsocontradicted the lint clause that rejects self-asserted confirmations. Writing the regression test surfaced a second defect: `by_key` was built once from disk, so two proposals sharing a key within a single drain both inserted — and a single drain processes a whole campaign's inbox.

**8. Cross-kernel confirmation was non-functional — including for the cards already committed.**
The writer now goes through `propose`, so `origin_kernels` is derived from `kernel_names` and present on new cards, and `rank` weights `confirms_blind` (double) rather than ignoring it.

The existing 135 cards are migrated rather than left behind. `kb.py backfill-origins` derives each card's provenance from `_inbox/*.json.drained` — the proposals `drain` already consumed, each carrying its run's `kernel_names` — taking theunion where a card was merged from several runs. Nothing is hand-written and nothing is guessed, since an invented origin would grant blind credit that was never earned. 130 of 135 are attributed; the remaining 5 give only an experiment directory or a lane name in `source:`, are listed by name in the command's output, and are left with no origin and no blind credit. Verified end to end: a backfilled card cited by a different kernel and winning now moves `confirms_blind` 0 → 1.

**Additional findings.** `deep_explore` now receives `LEARNED_KB` and is instructed to exclude `knowledge/learned/` when it is off. `kernel_workflow_bmk.js` forwards `use_learned_kb`. The generated index header now reports the per-class budget that actually binds, rather than a 40-line cap it was visibly exceeding. `lint --cards` exits non-zero on failure. The four standalone checks are now a CI step alongside `lint --cards` and `index --check` as gates — they are not pytest-collectable, since they assert at import and exit, which is why they had been left out.

Umangatamd (approved):

PR #415, opened 2026-08-19
Bench kernel candidates at the shapes the workload serves

### LINKED ISSUE #417: op_bench benches every candidate at max(m_buckets), but 94.1% of served passes are decode at M=64
## Summary

`op_bench.py` times every GEMM candidate at `max(m_buckets)` — always a prefill-scale shape — while **94.1% of served forward passes are decode at M=64**. The bake-off therefore picks winners on a shape the workload almost never runs. This is a measurement defect, not a kernel-quality problem, and it is a direct contributor to the "isolated win, no e2e gain" pattern.

Two sites, both on `main`:

* `e2e_workflow/scripts/op_bench.py:151` — `rep_m = max(buckets) if buckets else None`
* `e2e_workflow/scripts/op_bench.py:261` — `M = max(buckets)`

The docstring states the premise explicitly: *"Benched at the DOMINANT (largest) m_bucket — the GPU-time mass."* That premise is false. Mass is per-launch time x launches, and the largest bucket carries the fewest launches.

## Evidence — corpus-wide

Measured over `/shared_nfs/hyperloom-claw/`, 19 Aug 2026.

| measurement | value |
|---|---|
| decode share of served forward passes | **94.1%**, identical in all 679 observations |
| served decode M (`est_shape`) | **64**, in all 993 observations |
| served prefill M | 8192 / 16384 / 7211 / 2048 / 1024 |
| `opbench_result.json` files with no per-bucket timing | **122 of 123 (99%)** |
| of the 19 recording their bench M, benched at the decode shape | **0** |
| tasks named `*decode*` nonetheless benched at M=8192 | 3 |
| kernels whose `meta` spans >=2 serving regimes | 118 of 200 (59%) |

The three misbenched decode tasks: `aiter_a8w8_blockscale_ck_gemm_decode_task`, `decode_fp8_gemm_MT128x64_task`, `decode_fp8_gemm_MT32x32_task`.

**Natural experiment.** The one kernel in the campaign benched at a decode-scale grid (`kernel_paged_attention_2d`, `grid=(64,1)`) is also the largest win in the campaign — **9.472x**, and the corpus-highest Amdahl ceiling at **17.80%** (`MiniMax-M3-MXFP4/20260815T023841Z`).

## Evidence — measured on GPU

MI355X gfx950, `dense_linear_gemm_bf16_task`, 5 replicates, real `op_bench.py` from `main`. Full table and method: https://github.com/AMD-AGI/GEAK/pull/415#issuecomment-5339961179

The decision-changing case is `tunableop`, paired across replicates:

| shape | delta vs incumbent | t | verdict |
|---|---|---|---|
| M=8192 (what we bench) | +0.0047 ms, sd 0.0067 | **1.6** | indistinguishable |
| M=64 (94.1% of passes) | +0.0550 ms, sd 0.0013 | **96.5** | **1.89x slower** |

A deployable candidate that today's bench cannot separate from the incumbent is 1.89x slower on almost every forward pass. Served cost across backends is overstated by **4.8x to 9.6x**.

A second failure mode is invisible today: `flydsl` runs at decode and **fails at prefill**. The current bench reports it `available: false` and the partial-coverage fact is lost.

## Proposed invariant

> **Bench at the shapes the workload serves, and weight them by how often it serves them.**

Sweep the served buckets and blend to a single served-weighted `ms`. Credit a backend only where it ran correctly at *every* served shape. Publish `ms_by_bucket` so the blend is auditable.

## Scope note

This is one of three independent defects in the same chain, and fixing any one alone is not sufficient:

* shapes are guessed rather than measured — #384
* benched at the wrong shape — this issue
* the win is attributed to the wrong leg — #416

### LINKED ISSUE #418: op_bench publishes isolated_speedup 0.0 for bake-offs it never timed (318/318 attention results)
## Summary

`op_bench` publishes `isolated_speedup: 0.0` whenever no backend produced a timing. That is the same value it publishes for a bake-off that *was* timed and found nothing faster. The acceptance gate therefore cannot tell a kernel it declined from one it never benched.

`bench_attn` takes that path unconditionally. It returns one row with `available: True, correct: True, ms: None` — by design, because attention backend choice is a server-level flag (`--attention-backend`) delegated to the Config Tuner. The design decision is reasonable; publishing `0.0` for it is not.

## Evidence

All `opbench_result.json` under `/shared_nfs/hyperloom-claw`, sessions dated 14–19 Aug 2026. The `op_bench.py` staged in the session tree is byte-identical to `origin/main` (md5 `6ec52d11b75fc22bd4043a0e244411ac`), so this is the shipped code.

Splitting by producer — 40 rows carry keys `op_bench` never writes (`bench_method`, `dominant_bucket`, `m1_bucket`, `overlay_file`) and are hand-rolled harnesses in `_exp/team_*` workspaces that bypass `op_bench` entirely:

| produced by | op_kind | n | timed | isolated_speedup |
|---|---|---:|---:|---|
| `op_bench.py` | attn | **318** | **0** | `0.0` ×318 |
| `op_bench.py` | gemm | 87 | 69 | `0.0` ×21, `1.0` ×63, `1.0188` ×3 |
| hand-rolled bypass | attn | 14 | 14 | **`9.472` ×12**, `1.0` ×2 |
| hand-rolled bypass | moe | 23 | 0 | `1.6971` |
| hand-rolled bypass | fp8 gemm | 3 | 0 | `1.48` |

`op_bench` timed attention **0 times out of 318**. 90.7% of attention backend rows carry `ms: None`. `harness_suspect` was `false` on every one of the 318, because nothing raised — the guard requires an exception and this failure is silent.

The only real attention measurement in the campaign, **9.47×**, came from an agent that wrote its own bench keyed on `dominant_bucket` / `m1_bucket` — i.e. it hand-did what #415 makes systematic, and found a 9.47× win that `op_bench` could not see.

### The gate is blind where the gains are

23 sessions scored *every* attention bake-off `0.0`. Eight of them still produced a positive e2e result, arriving through the e2e A/B path rather than the isolated gate:

| session | attn bake-offs, all `0.0` | e2e those attention kernels delivered |
|---|---:|---|
| gemma-4-26B-A4B-it/20260815T130915Z | 52 | +18.915%, +14.973%, +10.512% |
| Kimi-K3/20260816T122327Z | 34 | +11.710% |
| Llama-3.1-8B-Instruct/20260816T114410Z | 24 | +2.901% |
| gemma-4-26B-A4B-it/20260814T155153Z | 18 | +10.383% |
| Qwen3-14B-FP8/20260816T050457Z | 15 | +7.530%, +6.779%, +0.705% |
| GLM-5.2-MXFP4/20260814T163244Z | 14 | +29.994%, +2.818% |
| gemma-4-26B-A4B-it/20260816T112750Z | 14 | +14.040% |
| Qwen3-14B-FP8/20260814T163051Z | 2 | +14.924% |

110.40 pct-points of best-per-session e2e gain, from sessions in which the isolated bench reported that nothing was faster.

Corroborating: across the same window `micro_speedup` correlates with realised e2e gain at Pearson **r = +0.64**, Spearman **ρ = +0.71** (n=21). The micro number transfers when it exists. The defect is that for attention it never exists, and its absence is encoded as a zero.

## The invariant

> `op_bench` must not publish an `isolated_speedup` it did not measure.

- No timing anywhere → `isolated_speedup: null`, `measured: false`.
- Timed, but no correct backend → `0.0` stands. That is a real verdict.
- A row asserting `available` **and** `correct` while carrying no `ms` contradicts itself → `harness_suspect`. A recorded skip (`available: False`) asserts nothing and stays out of the signal.

This is one rule over all op kinds. It does not special-case attention, and it does not make attention faster — it stops a kernel that was never benched from being scored as a kernel that lost.

## Replay of the fix

Every one of the 405 logged `op_bench` bake-offs, driven back through the real `main()` decision path with its recorded `results` list:

| op_kind | n | control | with the fix | changed |
|---|---:|---|---|---:|
| attn | 318 | `0.0` ×318, suspect 0 | `null` ×318, suspect 318 | 318 |
| gemm | 87 | `0.0` ×21, `1.0` ×63, `1.0188` ×3, suspect 18 | `null` ×18, `0.0` ×3, `1.0` ×63, `1.0188` ×3, suspect 18 | 18 |

Genuinely measured rows are untouched — the 63 at `1.0`, the 3 at `1.0188` and the 3 timed-but-incorrect `0.0` all keep their verdict. Only never-timed rows change.

## Scope

This issue is the *reporting* invariant only. Two related defects are deliberately **out of scope** and need their own issues:

1. **`bench_attn` still does not time attention.** With this fix it says so honestly instead of saying `0.0`. Actually timing the captured attention path at the served bucket is a larger change.
2. **`op_kind` routing.** `main()` routes on `op_kind == "gemm"`, so `moe` (23 results) and `fp8_a8w8_blockscale_gemm_bpreshuffle` (3) fall through to `bench_attn` and are never timed either. The invariant makes this visible rather than silent, but does not fix the routing.

iraj465:
Fixes #417
Fixes #418

## The defect

`op_bench.py` benched every candidate at one shape:

```python
M = max(buckets)   # "the DOMINANT (largest) m_bucket — the GPU-time mass"
```

Largest is not the mass. Mass is per-launch time × **launches**, and the analytic call model already sitting in `serving_weight_model.analytic_calls` counts `decode = OSL` passes against `prefill = CONC*ceil(ISL/chunk)`. Neither bucket dominates universally — prefill runs a few large passes, decode runs many small ones — so no single shape represents the mix, and the largest one is routinely the least-served.

`harness_lib.serving_weighted_speedup` has encoded the correct blend all along, but it has **no production Python caller**; it only runs if a role prompt is followed.

## Evidence

Corpus: `/shared_nfs/hyperloom-claw`, measured 19 Aug 2026.

| measurement | value |
|---|---|
| decode share of served forward passes | **94.1%**, identical in all 679 observations |
| served decode M (`est_shape`) | **64**, in all 993 observations |
| served prefill M | 8192 / 16384 / 7211 / 2048 / 1024 |
| `opbench_result.json` with no per-bucket timing | **122 of 123** |
| of the 19 recording their bench M, benched at the decode shape | **0** |
| tasks named `*decode*` benched at M=8192 | 3 |

The three: `aiter_a8w8_blockscale_ck_gemm_decode_task`, `decode_fp8_gemm_MT128x64_task`, `decode_fp8_gemm_MT32x32_task`.

## A/B replay over the same run logs

Replayed by calling the real `served_buckets`, over all 200 `kernels/*/meta.json`:

| | before | after |
|---|---|---|
| kernels benched at 1 shape | 60 | 0 |
| kernels swept over both served shapes | 0 | **60** |
| where the benched shape was the whole served mix | 0 of 60 | 60 of 60 |
| **median share of served passes the bench shape did not represent** | **94.1%** | 0% |

Old shape → shapes now swept:

| old | now swept | n |
|---|---|---|
| M=8192 | decode M=64, prefill M=8192 | 36 |
| M=16384 | decode M=64, prefill M=1024 | 6 |
| M=16384 | decode M=64, prefill M=16384 | 5 |
| M=1024 | decode M=64, prefill M=1024 | 4 |
| M=16384 | decode M=64, prefill M=8192 | 3 |
| *(6 more)* | decode M≈64, prefill = the old shape | 6 |

What those kernels scored under the single-shape bench:

| group | n | `==1.0` | `>1.05x` | no result |
|---|---|---|---|---|
| **would now sweep 2 buckets** | 54 | **47** | 1 | 6 |
| unchanged (0–1 served buckets) | 69 | 12 | 2 | 55 |

47 of 54 scored *exactly* 1.0 — no backend beat the baseline at a shape the vendor library is already tuned for.

The one kernel in the corpus benched at the decode shape (`grid=(64,1)`) is also the only large win: `kernel_paged_attention_2d`, **9.472x**, the corpus-highest Amdahl ceiling at **17.80%**.

## What this replay does and does not prove

It proves the **targeting** change exactly, because bench shape is a deterministic function of `meta`. It does **not** prove new kernel physics: the per-bucket timings do not exist in the logs (that is the defect), so re-measuring needs GPUs. No e2e number is claimed here.

## The change

- `harness_lib.served_buckets(meta)` — maps the serving model to the shapes to bench, snapping each phase to the nearest profiled `m_bucket`, honouring `served_regimes`, de-duplicating, and returning `[]` when there is no serving model.
- `bench_gemm` sweeps those buckets and `_merge_served` collapses the sweep into one served-weighted per-launch `ms`:

  `ms_served = Σ(ms_bucket × calls_bucket) / Σ(calls_bucket)`

Because the merged `ms` is what downstream reads, winner selection, `isolated_speedup` and the Amdahl ceiling all become served-weighted **with no further change**. `ms_by_bucket` publishes the per-shape numbers for audit. A candidate that wins at one served shape and breaks at another is not selectable.

Metas with no serving model, or a single served bucket, keep their previous single-shape behaviour.

## Tests

19 added. 840 pass; the 2 failures in `test_bench_e2e_teardown_lookup.py` are pre-existing on `main` (verified by stashing).

The regression test states the case the fix exists for: a candidate 5x faster at the prefill bucket and 2x slower at decode is a net **loss** once weighted, and is no longer selected.

---

# Second commit: never publish a speedup that was not measured

`Fixes #418`. Same thesis as the shape fix — the published number must come from a real measurement of the real workload — applied to the other half of the problem.

## The defect

`op_bench` published `isolated_speedup: 0.0` whenever no backend produced a timing, which is the same value it publishes for a bake-off that *was* timed and found nothing faster. The gate could not tell a kernel it declined from one it never benched.

`bench_attn` takes that path unconditionally: one row, `available: True, correct: True, ms: None`, because attention backend choice is a server-level flag delegated to the Config Tuner. That delegation is fine. Publishing `0.0` for it is not.

## Evidence

All `opbench_result.json` under `/shared_nfs/hyperloom-claw`, sessions 14–19 Aug 2026. The `op_bench.py` staged in the session tree is byte-identical to `origin/main` (md5 `6ec52d11b75fc22bd4043a0e244411ac`), so this is the shipped code. 40 results carrying keys `op_bench` never writes (`bench_method`, `dominant_bucket`, `m1_bucket`) are hand-rolled harnesses that bypass it, and are separated out:

| produced by | op_kind | n | timed | isolated_speedup |
|---|---|---:|---:|---|
| `op_bench.py` | attn | **318** | **0** | `0.0` ×318 |
| `op_bench.py` | gemm | 87 | 69 | `0.0` ×21, `1.0` ×63, `1.0188` ×3 |
| hand-rolled bypass | attn | 14 | 14 | **`9.472` ×12**, `1.0` ×2 |
| hand-rolled bypass | moe | 23 | 0 | `1.6971` |
| hand-rolled bypass | fp8 gemm | 3 | 0 | `1.48` |

`op_bench` timed attention **0 times out of 318**, and `harness_suspect` was `false` on all 318 because nothing raised — the existing guard requires an exception, and this failure is silent.

Note the third row. The only real attention measurement in the campaign, **9.47×**, came from an agent that wrote its own bench keyed on `dominant_bucket` / `m1_bucket` — it hand-did what the first commit of this PR makes systematic, and found a 9.47× win `op_bench` could not see.

### Where the gains actually were

23 sessions scored *every* attention bake-off `0.0`. Eight still produced a positive e2e result, arriving through the e2e A/B path rather than the isolated gate:

| session | attn bake-offs, all `0.0` | e2e those attention kernels delivered |
|---|---:|---|
| gemma-4-26B-A4B-it/20260815T130915Z | 52 | +18.915%, +14.973%, +10.512% |
| Kimi-K3/20260816T122327Z | 34 | +11.710% |
| Llama-3.1-8B-Instruct/20260816T114410Z | 24 | +2.901% |
| gemma-4-26B-A4B-it/20260814T155153Z | 18 | +10.383% |
| Qwen3-14B-FP8/20260816T050457Z | 15 | +7.530%, +6.779%, +0.705% |
| GLM-5.2-MXFP4/20260814T163244Z | 14 | +29.994%, +2.818% |
| gemma-4-26B-A4B-it/20260816T112750Z | 14 | +14.040% |
| Qwen3-14B-FP8/20260814T163051Z | 2 | +14.924% |

110.40 pct-points of best-per-session e2e gain, from sessions where the isolated bench reported nothing was faster. Across the same window `micro_speedup` correlates with realised e2e gain at Pearson **r = +0.64**, Spearman **ρ = +0.71** (n=21): the micro number transfers when it exists, and for attention it never exists.

## The invariant

> `op_bench` must not publish an `isolated_speedup` it did not measure.

- No timing anywhere → `isolated_speedup: null`, `measured: false`.
- Timed, no correct backend → `0.0` stands. That is a real verdict.
- A row asserting `available` **and** `correct` with no `ms` contradicts itself → `harness_suspect`. A recorded skip (`available: False`) asserts nothing and stays out of the signal.

One rule over all op kinds, no attention special-case.

## A/B — every logged bake-off, replayed through the real decision path

The `results` list is what the backends actually produced on the GPU and is recorded verbatim in each `opbench_result.json`. Feeding it back through `main()` exercises the genuine winner/speedup/`harness_suspect` code in both versions, so this is a replay of the real predicate rather than a paraphrase of it.

| op_kind | n | control (`origin/main`) | this PR | rows changed |
|---|---:|---|---|---:|
| attn | 318 | `0.0` ×318, suspect 0 | `null` ×318, suspect 318 | **318** |
| gemm | 87 | `0.0` ×21, `1.0` ×63, `1.0188` ×3, suspect 18 | `null` ×18, `0.0` ×3, `1.0` ×63, `1.0188` ×3, suspect 18 | 18 |

**No genuinely measured row moves.** The 63 at `1.0`, the 3 at `1.0188` and the 3 timed-but-incorrect `0.0` all keep their verdict. Only never-timed rows change — 336 of 405.

## Tests

5 added (`TestUnmeasuredIsNotZero`), 845 pass. Two failures in `test_bench_e2e_teardown_lookup.py` are pre-existing on the base commit and unrelated.

## What this does not fix

Deliberately out of scope, each needing its own issue:

1. **`bench_attn` still does not time attention.** It now says so honestly instead of saying `0.0`. Timing the captured attention path at the served bucket is a larger change.
2. **`op_kind` routing.** `main()` routes on `op_kind == "gemm"`, so `moe` (23 results) and `fp8_a8w8_blockscale_gemm_bpreshuffle` (3) fall through to `bench_attn` and are never timed either. This PR makes that visible instead of silent; it does not reroute them.
3. **The journey records no kernel artifact.** All 13 integrated wins in the window carry `optimized_files: []`, `patch_path: null`, `target_file: null`, while authored kernels demonstrably exist on disk. That is `interface/run_e2e.py`, a different subsystem.
4. **Profile capture lands in prefill.** 40 re-captures across 22 of 29 sessions (76%), every one succeeding and re-run only because the window missed decode. That is `bench_e2e.sh`.

---

## A/B: the bench shape actually changes

Over the 37 `kernels/*/meta.json` written since 13 Aug in `/shared_nfs/hyperloom-claw`, calling the shipped `served_buckets` predicate rather than a paraphrase of it. **8 of 9 GEMM metas change bench shape.**

| task | `main` M | with this PR |
|---|---:|---|
| `aiter_gemm_a16w16_dense_bf16_decode_task` | 8192 | **64, 8192** |
| `aiter_gemm_a16w16_dense_task` | 8192 | 64, 8192 |
| `dense_fp8_blockscale_gemm_task` | 8192 | 64, 8192 |
| `dense_bf16_gemm_bundle_task` | **65536** | 64, 65536 |
| `Cijk_Alik_Bljk_..._UserArgs_MT128x` | 8192 | 64, 8192 |
| `Cijk_Alik_Bljk_..._UserArgs_MT256x` | **16384** | 64, 8192 |
| `Cijk_MT128x64x128_gate_up_proj_task` | 8192 | 64, 8192 |
| `dense_gemm_family_hipblaslt_Cijk_MT16x16x1024_task` | 8192 | 64, 8192 |
| `Cijk_hipBLASLt_dense_GEMM_cluster_task` | 8192 | 8192 — no serving model, unchanged |

A task named `..._decode_task` was benched at M=8192. Two tasks were benched at 16384 and 65536, shapes that are not served buckets at all.

### Scope this does *not* cover

- **No GPU re-measurement.** The shapes above are recomputed from logged metas; no workload was re-run. Kernel timings at the new shapes are not measured here.
- **Single-bucket path:** 0 of the 9 GEMM metas is single-regime, so that path is a **precondition**, not a present-day change. It is exercised by the 14 single-regime metas in the sample (13 `['decode']`, 1 `['prefill']`), which are attention tasks — and attention never produces a timing at all today (#418).

### DIFF
--- a/e2e_workflow/e2e_workflow.js
+++ b/e2e_workflow/e2e_workflow.js
@@ -476,7 +476,10 @@ const EXTRACT_OP_SCHEMA = obj({
 const OPBENCH_SCHEMA = obj({
   short_name: { type: 'string' }, op_kind: { type: 'string' }, provenance_ok: { type: 'boolean' },
   winner_backend: { type: 'string' }, winner_kind: { type: 'string' },
-  isolated_speedup: { type: 'number' }, winner_editable: { type: 'boolean' },
+  // null when nothing was timed -- a speedup is a measurement, and 0.0 read as
+  // 'benched, not faster'. `measured` is the discriminator; gate on it, not on the number.
+  isolated_speedup: { type: ['number', 'null'] }, measured: { type: 'boolean' },
+  winner_editable: { type: 'boolean' },
   best_known_ms: { type: 'number' },
   recommend_tier_c: { type: 'boolean' }, author_plan: arrObj, tuning_artifact: { type: 'string' },
   apply_env: { type: 'string' }, apply_flags: { type: 'string' }, code_patch: { type: 'string' },
--- a/e2e_workflow/roles/op_benchmarker.md
+++ b/e2e_workflow/roles/op_benchmarker.md
@@ -301,6 +301,7 @@ Return JSON:
   "winner_backend": "aiter|hipblaslt|triton|flydsl|ck|none",
   "winner_kind": "env|flag|patch|none",
   "isolated_speedup": 1.0,
+  "measured": true,
   "winner_editable": false,
   "best_known_ms": 0.0,
   "recommend_tier_c": false,
@@ -321,6 +322,10 @@ Return JSON:
   "reason": "the route decision: direct_light winner and/or which languages to author, with Amdahl headroom"
 }
 ```
+- `measured` / `isolated_speedup` — copy BOTH straight from `opbench_result.json`; never fill one in
+  yourself. `measured:false` means no backend produced a timing, and then `isolated_speedup` is `null`,
+  not `0.0`. `0.0` means the bake-off ran and nothing was faster — a different fact, and the acceptance
+  gate needs to tell them apart.
 - `gate:"have_winner"` — a direct_light (env/flag) winner is ready to integrate now.
 - `gate:"author_recommended"` — no direct win, but `author_plan` is non-empty: the orchestrator should
   run `kernel_workflow` per the plan and integrate the fastest authored result that beats `best_known_ms`.
--- a/e2e_workflow/scripts/harness_lib.py
+++ b/e2e_workflow/scripts/harness_lib.py
@@ -741,6 +741,56 @@ def served_regimes(meta):
     return {str(r).strip().lower() for r in sr if str(r).strip()}
 
 
+def served_buckets(meta):
+    """The M buckets this kernel actually SERVES, with their analytic per-wave pass counts.
+
+    `op_bench` benched at `max(m_buckets)`, on the premise that the largest bucket carries the
+    GPU-time mass. It does not. Mass = per-launch time x LAUNCHES, and the analytic call model
+    (`serving_weight_model.analytic_calls`) counts prefill = CONC*ceil(ISL/chunk) passes against
+    decode = OSL passes. Neither bucket dominates universally -- prefill runs few large passes,
+    decode runs many small ones -- so a single bucket cannot represent the served mix, and the
+    LARGEST one is routinely the least-served shape.
+
+    Decode's M is the concurrency (snapped to a graph capture size); prefill's is the chunk, or
+    ISL when the server does not chunk. Each is matched to the nearest profiled `m_buckets` entry
+    when one exists, so the bench runs on a shape the profile actually saw.
+
+    Returns [(phase, M, calls), ...] ordered by descending calls, de-duplicated by M. Returns []
+    when meta carries no serving model (offline / unit-test metas), which leaves callers on their
+    previous single-bucket behaviour."""
+    swm = ((meta or {}).get("workload") or {}).get("serving_weight_model") or {}
+    calls = swm.get("analytic_calls") or {}
+    if not calls:
+        return []
+    buckets = [int(b) for b in (meta.get("m_buckets") or [])
+               if str(b).strip().lstrip("-").isdigit()]
+    served = served_regimes(meta)
+    want = {"decode": swm.get("conc"),
+            "prefill": swm.get("prefill_chunk") or swm.get("isl")}
+    out = {}
+    for phase, n in calls.items():
+        ph = str(phase).strip().lower()
+        if served and ph not in served:
+            continue                      # a phase this kernel does not run in carries no weight
+        m, n = want.get(ph), _int_or_none(n)
+        m = _int_or_none(m)
+        if not m or not n or n <= 0:
+            continue
+        if buckets:
+            m = min(buckets, key=lambda b: abs(b - m))
+        prev = out.get(m)
+        if prev is None or n > prev[2]:
+            out[m] = (ph, m, n)
+    return sorted(out.values(), key=lambda t: -t[2])
+
+
+def _int_or_none(x):
+    try:
+        return int(x)
+    except (TypeError, ValueError):
+        return None
+
+
 def serving_weighted_speedup(per_case, meta, *, identity_eps=1e-4, geomean=True):
     """Centralized PRIMARY-metric weighting for the immutable unittests — replaces the per-kernel
     hand-rolled `_serving_calls`/`weight` blocks so every UT applies the SAME audited rule. Enforces:
--- a/e2e_workflow/scripts/harness_lib.py
+++ b/e2e_workflow/scripts/harness_lib.py
@@ -685,6 +685,8 @@ def amdahl_ceiling(pct_gpu, isolated_speedup):
     if p > 1.0:
         p /= 100.0
     p = min(max(p, 0.0), 1.0)
+    if isolated_speedup is None:
+        return 0.0        # no measurement -> no attributable ceiling; same answer as a non-win
     s = float(isolated_speedup)
     if s <= 0:
         return 0.0
--- a/e2e_workflow/scripts/op_bench.py
+++ b/e2e_workflow/scripts/op_bench.py
@@ -791,13 +791,24 @@ def main():
     ran_any = any(r.get("ms") for r in results)
     raised = [r for r in results if r.get("raised") or r.get("backend") == "ERROR"
               or "raised" in str(r.get("note", "")) or "failed:" in str(r.get("note", ""))]
-    harness_suspect = bool(results) and (not ran_any) and len(raised) > 0
+    # A row that claims it was AVAILABLE and CORRECT while carrying no ``ms`` contradicts
+    # itself: it reports a verdict on a timing it never took. Requiring an exception missed
+    # this quiet form, and the quiet form is the common one. A recorded skip
+    # (``available: False``) asserts nothing and stays out of the fault signal.
+    unmeasured_claim = [r for r in results
+                        if r.get("available") and r.get("correct") and r.get("ms") is None]
+    harness_suspect = bool(results) and (not ran_any) and bool(raised or unmeasured_claim)
     harness_error = ""
     if harness_suspect:
-        r0 = raised[0]
-        harness_error = str(r0.get("note") or r0.get("trace") or "unknown harness error")[:400]
-    speedup = (baseline["ms"] / winner["ms"]) if (winner and baseline and winner["ms"]) else (
-        1.0 if winner else 0.0)
+        r0 = (raised or unmeasured_claim)[0]
+        harness_error = str(r0.get("note") or r0.get("trace")
+                            or "backend reported available+correct but produced no timing")[:400]
+    # A speedup is a MEASUREMENT, so absent a measurement there is no number to publish.
+    # Reporting 0.0 for "never timed" made it indistinguishable from "timed, not faster",
+    # and the acceptance gate cannot tell a kernel it declined from one it never ran.
+    measured = bool(ran_any)
+    speedup = ((baseline["ms"] / winner["ms"]) if (winner and baseline and winner["ms"])
+               else (1.0 if winner else (0.0 if measured else None)))
     wb = winner["backend"] if winner else None
     # Only triton/hip are source-editable (-> Tier-C kernel-squad rewrite). ck is a library backend.
     editable = bool(wb in ("triton", "hip"))
@@ -838,7 +849,8 @@ def main():
         "winner_ms": winner["ms"] if winner else None,
         "baseline_backend": baseline["backend"] if baseline else None,
         "baseline_ms": baseline["ms"] if baseline else None,
-        "isolated_speedup": round(speedup, 4),
+        "measured": measured,
+        "isolated_speedup": round(speedup, 4) if speedup is not None else None,
         "pct_gpu_time": pct_gpu,
         "amdahl_ceiling_e2e_pct": amdahl_ceiling_pct,
         "winner_editable": editable,
--- a/e2e_workflow/scripts/op_bench.py
+++ b/e2e_workflow/scripts/op_bench.py
@@ -745,9 +745,15 @@ def bench_attn(args, meta):
     if not os.path.exists(iopath):
         return [{"backend": "current", "available": False, "correct": False, "ms": None,
                  "note": "attn bake-off needs reference_io.pt (captured q/k/v/meta); none found"}]
+    # A recorded SKIP, not a verdict. This path takes no timing at all, so `available: True,
+    # correct: True, ms: None` claimed a result it never measured -- which is exactly the
+    # self-contradiction main() now flags. Attention is not a contradictory row; it is a deliberate
+    # delegation, so it says so and stays out of the fault signal. This keeps the shipped attention
+    # cards true (`harness_suspect=false expected`) and leaves that rule for rows that really do
+    # contradict themselves.
     note = ("attention backend comparison is a SERVER-level flag (--attention-backend) -> delegated to "
-            "the Config Tuner fast path; op-level here only validates the oracle")
-    return [{"backend": "current", "available": True, "correct": True, "ms": None,
+            "the Config Tuner fast path; op-level bake-off skipped (no timing taken here)")
+    return [{"backend": "current", "available": False, "correct": False, "ms": None,
              "note": note, "artifact": iopath}]
 
 
--- a/e2e_workflow/scripts/op_bench.py
+++ b/e2e_workflow/scripts/op_bench.py
@@ -437,9 +437,15 @@ def _merge_served(per_bucket):
 
 def bench_gemm(args, meta):
     """Bake-off across the SERVED buckets, not just the largest one. Falls back to the historical
-    single-bucket bench when meta carries no serving model, or names only one served shape."""
+    single-bucket bench only when meta names no served shape at all.
+
+    One served bucket is still a served shape, and it is the case that matters most: a decode-only
+    kernel resolves to exactly one bucket, and falling back there benched it at ``max(m_buckets)``
+    -- a prefill shape the kernel never runs. ``_merge_served`` over a single bucket is the identity
+    on ``ms``, so routing one and many down the same path needs no special case and gains the
+    ``ms_by_bucket`` audit trail."""
     buckets = _hlib.served_buckets(meta) if _hlib is not None else []
-    if len(buckets) < 2:
+    if not buckets:
         return _bench_gemm_at(args, meta)
     return _merge_served([(ph, M, n, _bench_gemm_at(args, meta, M)) for ph, M, n in buckets])
 
--- a/e2e_workflow/scripts/op_bench.py
+++ b/e2e_workflow/scripts/op_bench.py
@@ -139,16 +139,19 @@ def _dtype(torch, name):
             }.get(str(name).lower(), torch.bfloat16)
 
 
-def _resolve_shape(shape, meta):
+def _resolve_shape(shape, meta, m_override=None):
     """Resolve a shape that may carry SYMBOLIC dims (e.g. "M" for a dynamic GEMM row count) into
     concrete ints. String dims are mapped via meta: "M"/"m*"/"-1"/None -> a representative value from
-    meta["m_buckets"] (the dominant = LARGEST profiled bucket). Raises a CLEAR error if a symbolic dim
+    meta["m_buckets"] -- `m_override` when the caller is benching a specific served bucket, else the
+    largest profiled one. Raises a CLEAR error if a symbolic dim
     cannot be resolved, so the caller records a meaningful harness error instead of a cryptic
     `randn(str, int, ...)` TypeError (the exact bug this guards against)."""
     if not shape:
         raise ValueError("empty/None shape")
     buckets = [int(b) for b in (meta.get("m_buckets") or []) if str(b).strip().lstrip("-").isdigit()]
-    rep_m = max(buckets) if buckets else None
+    # `m_override` is the bucket the caller is currently benching (see bench_gemm's served sweep).
+    # Without one, fall back to the largest profiled bucket.
+    rep_m = int(m_override) if m_override else (max(buckets) if buckets else None)
     out = []
     for d in shape:
         if isinstance(d, bool):
@@ -247,18 +250,18 @@ def _synth_blockscale_case(torch, meta, M, device, seed):
     return {"x": x_q, "w": w_q, "x_scale": x_scale, "w_scale": w_scale, "ref": ref, "M": M, "out_dt": out_dt}
 
 
-def bench_blockscale_gemm(args, meta):
+def bench_blockscale_gemm(args, meta, m_override=None):
     """Bake-off for an fp8 a8w8 blockscale GEMM head. The generic dense torch-BLAS backends cannot
     represent this op (fp8 + per-block scales), so the candidates are the meta callable(s):
     the live baseline blockscale path + its bpreshuffle variant (same signature
-    fn(x, w, x_scale, w_scale, dtype=out)). Benched at the DOMINANT (largest) m_bucket — the GPU-time
-    mass. Returns the standard per-backend results list."""
+    fn(x, w, x_scale, w_scale, dtype=out)). Benched at `m_override` — the served bucket the caller is
+    sweeping — else the largest profiled bucket. Returns the standard per-backend results list."""
     torch = _torch()
     device = "cuda" if torch.cuda.is_available() else "cpu"
     buckets = [int(b) for b in (meta.get("m_buckets") or []) if str(b).strip().lstrip("-").isdigit()]
     if not buckets:
-        buckets = [int(_resolve_shape(meta.get("a_shape") or ["M"], meta)[0])]
-    M = max(buckets)
+        buckets = [int(_resolve_shape(meta.get("a_shape") or ["M"], meta, m_override)[0])]
+    M = int(m_override) if m_override else max(buckets)
     case = _synth_blockscale_case(torch, meta, M, device, args.seed)
     out_dt = case["out_dt"]
 
@@ -309,7 +312,7 @@ def bench_blockscale_gemm(args, meta):
     return results
 
 
-def _load_or_synth_gemm(torch, task, meta, device, seed):
+def _load_or_synth_gemm(torch, task, meta, device, seed, m_override=None):
     """Return (A, B, bias, transpose_b, ref). Prefer the recorded oracle; else synthesize + compute ref
     with the default backend (perf is value-independent; this only fixes the correctness target)."""
     dt = _dtype(torch, meta.get("dtype", "bf16"))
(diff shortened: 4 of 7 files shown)


iraj465:
## Measured on GPU — MI355X gfx950, `dense_linear_gemm_bf16_task`

The log replay in the PR body proves the **targeting** change only. Per-bucket timings do not exist in the logs — that is the defect. So I re-measured: real `op_bench.py` from `origin/main` vs this branch, same task, same node, 5 replicates.

Task: `a=[M,4096] b=[28672,4096]` bf16, `m_buckets [8192,64,1]`, served mix `{prefill:64, decode:1024}` → **94.1% of forward passes at M=64**.

### The bench shape hides the cost

| backend | main: ms @M=8192 | new: @M=64 (decode) | new: @M=8192 (prefill) | new: served-weighted | main overstates by |
|---|---|---|---|---|---|
| hipblaslt | 1.2668 | 0.0618 | 1.2664 | 0.1327 | **9.6x** |
| rocblas | 1.2750 | 0.0618 | 1.2662 | 0.1327 | 9.6x |
| tunableop | 1.2830 | 0.1168 | 1.2711 | 0.1847 | 7.0x |
| ck | 2.0973 | 0.1211 | 2.0903 | 0.2369 | 8.9x |
| aiter | 2.8071 | 0.4503 | 2.8211 | 0.5897 | 4.8x |
| flydsl | *unavailable* | 0.1678 | **fails** | *refused — incomplete coverage* | — |

### Why it changes decisions

Penalty of picking the wrong backend, vs hipblaslt:

| backend | at the bench shape M=8192 | at the served shape M=64 | understated by |
|---|---|---|---|
| tunableop | **1.00x** | **1.89x** | 1.9x |
| ck | 1.65x | 1.96x | 1.2x |
| aiter | 2.23x | 7.28x | **3.3x** |

`tunableop` is the case that matters. Paired over 5 replicates:

* at M=8192: `+0.0047 ms`, sd 0.0067, **t=1.6** — indistinguishable from the incumbent.
* at M=64: `+0.0550 ms`, sd 0.0013, **t=96.5** — **1.89x slower**, on 94.1% of passes.

Today's bench cannot tell these apart, and `tunableop` is *deployable* (it emits a CSV). A candidate that is free at the benched shape and 1.89x slower at the served shape is precisely the "isolated win, no e2e gain" signature.

`flydsl` is the second case: it runs at decode and **fails at prefill**. Main reports it `available: false` and the failure is invisible. The branch measures it and refuses credit for not covering the served mix.

### What this does NOT show

On this task the branch also flipped the winner hipblaslt→rocblas in 2 of 5 runs (`isolated_speedup` 1.0031x). **That is noise, not a result** — paired hipblaslt−rocblas is `0.00000 ± 0.00039 ms` at decode. This kernel has no real win available, and both arms agree on that. The fix corrects the *measurement*; it does not manufacture a win.

Replication quality: prefill CV 0.09–0.47%, decode CV 0.12–2.4% (aiter 13.7%, the one noisy backend). n=5.

yueliu14:
hi @iraj465  Could you provide some end2end workflow results on workload? Did you see any changes of measuered shapes in e2e workflow with/without this PR?

yueliu14:
hi @iraj465  Here are comments of this PR:

1. Could you provide some end2end workflow results on workload? Did you see any changes of measuered shapes in e2e workflow with/without this PR?

2. Math checks out, and the coverage invariant is properly enforced — `_merge_served` ANDs `correct` across buckets, so a candidate that breaks at one served shape can't win. One gap:

**A single-regime kernel computes the right M and then discards it.**

```python
if len(buckets) < 2:
    return _bench_gemm_at(args, meta)      # m_override=None -> max(m_buckets)
```

For a decode-only kernel, `served_regimes` filters prefill out and `served_buckets` returns exactly one entry. Verified: `served_buckets -> [('decode', 64, 1024)]` → `m_override=None` → `M=8192`.

The commit message frames one-bucket metas as intentionally keeping previous behaviour, but for these kernels previous behaviour is the bug — the three `*decode*` tasks cited in the message (benched at M=8192) are single-regime and still land on 8192 after this change. `_merge_served` degenerates cleanly at n=1, so dropping the short-circuit costs nothing; the accompanying `test_a_single_served_bucket_is_not_wrapped_in_a_sweep` pins the current path.

Related: `m = min(buckets, key=lambda b: abs(b - m))` snaps the computed M back onto a profiled bucket. `kernel_extractor.md:118` notes decode M is routinely missing from the capture window, so where the profile missed decode the snap returns the shape the PR is trying to move away from. Worth preferring the computed M when no profiled bucket is close, rather than snapping unconditionally.

iraj465:
Thanks @yueliu14 — both points taken. The second is a real bug; fixed in `a8edd86`.

### 2. Single-regime kernel discards the M it computed — fixed

You are right that "previous behaviour" is exactly the defect for these kernels. `_merge_served` over one bucket is the identity on `ms`, so the special case was never needed to keep the number intact — I deleted it rather than adding a branch:

```python
buckets = _hlib.served_buckets(meta) if _hlib is not None else []
if not buckets:                      # was: if len(buckets) < 2
    return _bench_gemm_at(args, meta)
return _merge_served([(ph, M, n, _bench_gemm_at(args, meta, M)) for ph, M, n in buckets])
```

Only a meta naming *no* served shape falls back now. One and many take the same path, and the single-bucket case gains the `ms_by_bucket` audit trail. The test that asserted `calls == [None]` encoded the bug; it now asserts `[64]` for decode-only and `[8192]` for prefill-only, plus a third test that the timing is unperturbed.

### 1. Measured shapes, with and without the PR

Over the 37 `kernels/*/meta.json` written since 13 Aug in `/shared_nfs/hyperloom-claw`, calling the real `served_buckets` (not a paraphrase):

**GEMM tasks — 8 of 9 change bench shape**

| task | `main` M | with PR |
|---|---:|---|
| `aiter_gemm_a16w16_dense_bf16_**decode**_task` | 8192 | **64, 8192** |
| `aiter_gemm_a16w16_dense_task` | 8192 | 64, 8192 |
| `dense_fp8_blockscale_gemm_task` | 8192 | 64, 8192 |
| `dense_bf16_gemm_bundle_task` | **65536** | 64, 65536 |
| `Cijk_..._UserArgs_MT128x` | 8192 | 64, 8192 |
| `Cijk_..._UserArgs_MT256x` | **16384** | 64, 8192 |
| `Cijk_MT128x64x128_gate_up_proj_task` | 8192 | 64, 8192 |
| `dense_gemm_family_hipblaslt_Cijk_MT16x16x1024_task` | 8192 | 64, 8192 |
| `Cijk_hipBLASLt_dense_GEMM_cluster_task` | 8192 | 8192 *(no serving model — unchanged)* |

A task named `..._decode_task` was being benched at M=8192. Two tasks were benched at shapes (16384, 65536) that are not served buckets at all.

**On the single-bucket path specifically — honest scope:** 0 of these 9 GEMM metas is single-regime, so your bug does not fire on today's GEMM traffic. It fires on the 14 single-regime metas in the sample (13 `['decode']`, 1 `['prefill']`), 12 of which resolve to exactly one bucket. Those are all attention tasks, and attention today routes to `bench_attn`, which never produces a timing at all (#418). So this fix is a **precondition** for those kernels rather than an immediate change in their numbers — worth stating plainly rather than claiming a win it does not yet deliver.

**No e2e workflow re-run yet.** I have no GPU held right now (queued on `QOSGrpNodeLimit`), so everything above is computed from the logged metas by calling the shipped predicate, not from a fresh workload run. I will post the e2e numbers once a node lands.

yueliu14:
Second commit — the evidence here is strong, and the replay A/B (feeding recorded `results` lists back through the real `main()`) is a much better methodology than a paraphrased predicate. The `0.0` collision is real and the 9.47× found by a hand-rolled bench that bypassed op_bench makes the cost concrete. Two things the change collides with:

**1. `null` fails the result schema.**

`e2e_workflow.js:479,486`:

```js
isolated_speedup: { type: 'number' }, ...
}, ['gate', 'isolated_speedup']);   // required
```

Same shape in `kernel_workflow.js:113,121`. `isolated_speedup: null` is both wrong-typed and present-but-null against a required key, so the bake-off result won't validate unless the schemas move with it. `harness_lib.amdahl_ceiling` also does `float(isolated_speedup)` (`:688`) — worth confirming the null never reaches it.

**2. Flipping all 318 attn rows to `harness_suspect` contradicts shipped knowledge cards.**

`harness_suspect` drives a branch downstream (`e2e_workflow.js:1649`, `:1847`):

```js
const harness = !!(bake && (bake.gate === 'harness_error' || bake.harness_suspect));
```

and three learned cards currently teach the opposite — `paged-attn-nonpow2-gfx942.md:19` says "harness_suspect=false is expected, NOT a fault", and `:51` "bench_attn only validates the oracle (harness_suspect=false expected, rel=0)". Per the A/B table this PR turns that into `suspect 318`, which trades 318 silent zeros for 318 false alarms and needs the cards updated in lockstep.

The commit's own invariant already has the better slot for this: a recorded skip (`available: False`) "asserts nothing and stays out of the signal". Attention isn't a self-contradicting row — it's a deliberate delegation to the Config Tuner. Having `bench_attn` record a skip rather than `available: True, correct: True, ms: None` gives `measured: false, harness_suspect: false`, keeps the cards true, and leaves the self-contradiction rule for rows that are genuinely self-contradicting.

iraj465:
Thanks @yueliu14 — both confirmed, both fixed in `62538e6`. Branch also rebased onto `main`.

### 2. `bench_attn` records a skip

Your reading is right, and it is the better fix. The row was `available: True, correct: True, ms: None` while taking no timing at all — the self-contradiction the commit had just learned to flag. Attention is not a contradictory row; it is a delegation. So it says so:

```python
return [{"backend": "current", "available": False, "correct": False, "ms": None,
         "note": "...delegated to the Config Tuner fast path; op-level bake-off skipped "
                 "(no timing taken here)", "artifact": iopath}]
```

The invariant already had the slot, as you said — a recorded skip asserts nothing and stays out of the signal. No new branch, no card edits: `paged-attn-nonpow2-gfx942.md:19,51` and the other three cards stay true.

**Three-arm replay, all 405 logged bake-offs, 14–19 Aug, through the real `main()`.** Arms 1–2 inject the recorded `results`. Arm 3 does too for GEMM, but for attention calls the **shipped new `bench_attn`** — an injected row is the *old* producer's output and can never show a producer fix.

| op_kind | n | control | `a8edd86` | with `62538e6` |
|---|---:|---|---|---|
| attn | 318 | `0.0` ×318, suspect **0** | `null` ×318, suspect **318** | `null` ×318, suspect **0** |
| gemm | 87 | `0.0` ×21, `1.0` ×63, `1.0188` ×3, suspect 18 | `0.0` ×3, `null` ×18, `1.0` ×63, `1.0188` ×3, suspect 18 | unchanged |

318 silent zeros go without becoming 318 false alarms. The 18 real GEMM harness faults still fire; no measured row moves.

> Two false starts worth naming, since both produced clean-looking wrong numbers. A stale monkeypatch let one file's `results` answer for the next; and with no `torch` on the box, `main()` caught the `ImportError` into a `backend: "ERROR"` row, so arm 3 first reported suspect 318 — the import failing, not the fix.

### 1. `null` vs the schema

Confirmed at `e2e_workflow.js:479,486` and `kernel_workflow.js:113,121` — wrong-typed *and* required. The schemas now say what the field means, using the `['string','null']` form already in `kernel_lane.js:284`:

```js
isolated_speedup: { type: ['number', 'null'] }, measured: { type: 'boolean' },
```

`op_benchmarker.md` gets the matching contract line: copy both from `opbench_result.json`, never fill one in; `null` ≠ `0.0`.

On `amdahl_ceiling` — it is **already unreachable** with a null: `op_bench.py:844` guards on `and winner`, and where a winner exists `speedup` is `baseline/winner.ms` or `1.0`, never `None`. Guarded anyway, since it is a public helper and `harness_lib.py:708` also calls it:

```python
if isolated_speedup is None:
    return 0.0        # no measurement -> no attributable ceiling; same answer as a non-win
```

### Tests

1016 pass. New: the real `bench_attn` row is unmeasured **and** not a fault; `amdahl_ceiling(0.5, None) == 0.0`. The old `bench_attn` test asserted `available/correct == True`, so it encoded the bug — replaced, not adjusted.

The 2 failures in `test_bench_e2e_teardown_lookup.py` are pre-existing; they fail on a detached checkout of current `main` too.

### On your first question

No e2e workflow run yet — no GPU held. Everything above is the real decision path replayed over logged artifacts. The on-GPU MI355X numbers in the earlier comment cover the targeting change only.

yueliu14 (approved):

Umangatamd (approved):

PR #422, opened 2026-08-19
fix(bench_e2e): make the shared serving-GPU lock openable, and its failure loud

### LINKED ISSUE #421: bench_e2e: shared /tmp serving-GPU lock is unopenable by a second user, and the refusal is silent
## Summary

`bench_e2e.sh` serializes every serving launch behind one lock file per GPU set:

```bash
SERVING_LOCK="${SERVING_GPU_LOCK:-/tmp/geak_serving_gpu_${_gpu_key}.lock}"
exec {SERVING_LOCK_FD}>"$SERVING_LOCK"
echo ">>> Acquiring serving-GPU lock ($SERVING_LOCK) for GPU=$GPU ..."
if ! flock -w "${SERVING_LOCK_WAIT:-7200}" "$SERVING_LOCK_FD"; then
```

Sharing the file is **correct** — the GPUs are shared, so a per-user lock would hand
two users the same cards. But the file is created with the creator's umask, and
nothing in the repo ever sets `SERVING_GPU_LOCK`. The second user on a host cannot
open it.

## The failure is silent

`exec {FD}>"$LOCK"` does **not** stop the shell when the open fails. Reproduced
exactly:

```
line 3: /tmp/geak_serving_gpu_0.lock: Permission denied   <- stderr only
>>> Acquiring serving-GPU lock (...) for GPU=0 ...        <- stdout, prints anyway
line 5: SERVING_LOCK_FD: unbound variable                 <- set -u ends the script
```

`flock` never runs. The reassuring line prints; the line naming the cause goes to
stderr, which the caller does not read. To anything watching stdout, the launch
simply never happened.

On this host:

```
-rw-r--r-- 1 <other-user> ubuntu 0 Aug 19 10:14 /tmp/geak_serving_gpu_0.lock
```

## Where it does and does not bite

Measured, not assumed, over the Hyperloom claw campaign:

| measure | n |
|---|---:|
| GEAK log files scanned | 2000 |
| logs carrying the `Acquiring serving-GPU lock` line | 816 |
| acquire attempts | 917 |
| `serving-GPU lock acquired` confirmations | 912 |
| explicit `serving-GPU lock timeout` | 2 |
| **permission denials on the lock** | **0** |
| **`SERVING_LOCK_FD: unbound variable` deaths** | **0** |

*(Counts revised upward after a full sweep. An earlier pass globbed only
`e2e_cycle*/logs/*.log`, which reaches 1094 of the 2000 GEAK logs and misses
`baseline/`, `final/bench/`, `validation/` and `config/` — where this script also
runs. The one log matching both "Permission denied" and "lock" is `py-spy` failing
to attach, not the lock. The zeros are unchanged over the full set.)*

**The campaign is unaffected** — it runs single-tenant, so the creator is always the
opener. This is a shared-host and developer-machine defect, and I am filing it with
that scope stated rather than dressed up.

It is not hypothetical there: it reproduces today as **2 failures in this repo's own
test suite** on any host where another user created the lock first.

```
FAILED test_bench_e2e_teardown_lookup.py::test_contract_staged_beside_the_copy_passes_the_gate
FAILED test_bench_e2e_teardown_lookup.py::test_skill_dir_fallback_is_used_and_announced
AssertionError: 'STUB_LAUNCH_REACHED' not found in '... >>> Acquiring serving-GPU lock ...'
```

Those two tests also reveal a second problem: they take the machine's **real** GPU
mutex from a unit test. On a busy shared box that blocks for `SERVING_LOCK_WAIT`
(default 7200s) rather than failing.

## Fix

Keep the lock shared — that is the point of it — and make sharing actually work:
create it world-writable, test the open, and exit with a message that names the
owner, the mode, and both ways out. Give the unit tests their own lock path.

Not `${USER}` in the lock name: that would give two users disjoint locks on the same
GPUs and quietly remove the mutex, which is worse than the bug.

iraj465:
Closes #421

## The defect

One lock file per GPU set, `/tmp/geak_serving_gpu_<gpu>.lock`, and nothing in the
repo ever sets `SERVING_GPU_LOCK`. Sharing it is correct — the GPUs are shared, so a
per-user lock would hand two users the same cards. But the file is created with the
creator's umask, so the second user on a host cannot open it, and the refusal is
silent: `exec {FD}>"$LOCK"` does not stop the shell, so the script prints
`>>> Acquiring serving-GPU lock ...` with no fd assigned and then dies on the next
line with `SERVING_LOCK_FD: unbound variable`. `flock` never runs.

## A/B

**Before / after, same host, same suite.**

| test | before | after |
|---|---|---|
| `test_contract_staged_beside_the_copy_passes_the_gate` | **FAIL** — `'STUB_LAUNCH_REACHED' not found` | pass |
| `test_skill_dir_fallback_is_used_and_announced` | **FAIL** — same | pass |
| `test_an_unopenable_lock_fails_loudly_instead_of_silently` *(new)* | **FAIL** | pass |
| `test_a_shared_lock_is_left_openable_by_the_next_user` *(new)* | **FAIL** | pass |
| repo suite (`--ignore=examples`) | 1077 passed, **2 failed** | **1079 passed, 0 failed** |

Both new tests were run against unpatched `bench_e2e.sh` to confirm they fail there —
a test that passes before the fix proves nothing.

**Behaviour of the failure path:**

| | before | after |
|---|---|---|
| stdout | `>>> Acquiring serving-GPU lock (...)` then nothing | same line, then the refusal |
| stderr | `Permission denied` + `SERVING_LOCK_FD: unbound variable` | `cannot open serving-GPU lock (...): -rw-r--r-- 1 <owner> ...` + both remedies |
| exit code | 1 (unbound variable) | **4** (the script's own lock-failure code) |
| `flock` called | no | no — but now on purpose, and it says so |

**Campaign impact — stated honestly:**

| measure over the claw campaign | n |
|---|---:|
| GEAK log files scanned | 2000 |
| logs carrying the lock line | 816 |
| acquire attempts / confirmations | 917 / 912 |
| explicit lock timeouts | 2 |
| **permission denials** | **0** |
| **`SERVING_LOCK_FD: unbound variable` deaths** | **0** |
| **silent `unbound variable` deaths** | **0** |

The campaign runs single-tenant, so it never hits this. **No campaign run was lost to
it, and this PR recovers no e2e gain.** It fixes a shared-host and developer-machine
defect that today fails this repo's own suite.

## What changed

1. **Create the lock under `umask 000`** so the next user can open it. Best-effort —
   a single-tenant box never needs it, so a failure to widen is not fatal.
2. **Open with `>>` and test the open.** Append rather than truncate (`>` on a lock
   another process holds gains nothing), and exit 4 with a message naming the owner,
   the mode, and both ways out (`SERVING_GPU_LOCK`, `SERVING_GPU_LOCK_DISABLE`).
3. **Give the two unit tests their own lock path.** They ran past the gate and took
   the machine's *real* GPU mutex; on a busy box that blocks for `SERVING_LOCK_WAIT`
   (7200s) instead of failing.

## What I deliberately did not do

Put `${USER}` or `$(id -u)` in the lock name. It makes the symptom vanish and removes
the mutex: two users would hold disjoint locks over the same GPUs and benchmark on
top of each other. The lock is shared because the hardware is.

### DIFF
--- a/e2e_workflow/scripts/bench_e2e.sh
+++ b/e2e_workflow/scripts/bench_e2e.sh
@@ -355,8 +355,25 @@ trap server_teardown EXIT
 if [ "${SERVING_GPU_LOCK_DISABLE:-0}" != "1" ] && [ "${REUSE_SERVER:-0}" != "1" ]; then
   _gpu_key="${GPU:-0}"; _gpu_key="${_gpu_key//,/_}"
   SERVING_LOCK="${SERVING_GPU_LOCK:-/tmp/geak_serving_gpu_${_gpu_key}.lock}"
-  exec {SERVING_LOCK_FD}>"$SERVING_LOCK"
+  # The lock file is SHARED on purpose -- the GPUs are shared, so a per-user lock would hand two
+  # users the same cards and defeat the mutex. Shared means whoever creates it first must leave it
+  # openable by the next user, hence the group/other write bit. Failure to widen it is not fatal
+  # (a single-tenant box never needs it); failure to OPEN it is.
+  if [ ! -e "$SERVING_LOCK" ]; then
+    (umask 000; : > "$SERVING_LOCK") 2>/dev/null || true
+  fi
   echo ">>> Acquiring serving-GPU lock ($SERVING_LOCK) for GPU=$GPU ..."
+  # Append, never truncate: `>` on a lock another process holds gains nothing and needs the same
+  # write bit. And this open MUST be tested. A failed `exec {FD}>` does not stop the shell, so the
+  # script used to print "Acquiring ..." with no fd assigned, then die on the next line with
+  # `SERVING_LOCK_FD: unbound variable` under `set -u` -- flock never ran, and the only line naming
+  # the cause went to stderr, which the caller does not read.
+  if ! exec {SERVING_LOCK_FD}>>"$SERVING_LOCK"; then
+    echo "!!! cannot open serving-GPU lock ($SERVING_LOCK): $(ls -ld "$SERVING_LOCK" 2>&1)" >&2
+    echo "!!! it is shared across users by design; make it group/other writable, or set" >&2
+    echo "!!!   SERVING_GPU_LOCK=<path> (a different lock)  /  SERVING_GPU_LOCK_DISABLE=1 (no mutex)" >&2
+    exit 4
+  fi
   if ! flock -w "${SERVING_LOCK_WAIT:-7200}" "$SERVING_LOCK_FD"; then
     echo "!!! serving-GPU lock timeout (${SERVING_LOCK_WAIT:-7200}s) on GPU=$GPU" >&2
     exit 4
--- a/e2e_workflow/scripts/tests/test_bench_e2e_teardown_lookup.py
+++ b/e2e_workflow/scripts/tests/test_bench_e2e_teardown_lookup.py
@@ -15,7 +15,9 @@ start it -- and that refusal is on the launch path of EVERY server in the produc
 it is asserted here rather than trusted.
 
 The gate sits before the serving-GPU lock and before any launch, so these tests need
-neither a GPU nor a model: they assert the exit code and the stderr remedy.
+neither a GPU nor a model: they assert the exit code and the stderr remedy. The two that
+run PAST the gate do reach the lock, so they are given their own lock path -- a unit test
+must not contend for the host's real GPU mutex.
 """
 import os
 import shutil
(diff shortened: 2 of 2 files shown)


iraj465:
Closing. The defect is real and reproduces locally, but it never fired in the campaign: across all 2000 GEAK logs there are 0 permission denials and 0 `SERVING_LOCK_FD: unbound variable` deaths, because the campaign runs single-tenant. Not worth reviewer time at that impact. The branch `fix/serving-lock-shared-tmp` stays pushed if anyone hits this on a shared host.

PR #431, opened 2026-08-25
Merge/warmup plus tuning ci fixed

amd-ethany:
## What this is

Merge of two feature branches — `feat/kernel-warm-start-kb-artifacts` and `feat/tuning-skillset` — plus the Hyperloom measurement alignment that both depend on. 34 commits, ~26k insertions.

The two features answer the same complaint from different directions: **GEAK re-derived the same tuning from scratch on every run, and had no vocabulary for "tune the existing kernel" as distinct from "author a new one."** The KB gives a run access to what previous runs found; the skillset gives it a procedure to follow when there is nothing to recall.

---

### 1. Warm-start knowledge base (`kb/`, `*_store.py`)

A run can now open with what earlier runs on the same workload already established, instead of an empty context.

- **`kb/` package** — `store_local.py` / `store_remote.py` / `store_client.py` behind one interface, so a lane reads the same whether the entry came from a directory or from behind a KB Store key. Plus `identity.py` (content signatures), `attest.py`, `curate.py`, `ladder.py`, `retract.py`.
- **`kernel_workflow/scripts/experience_store.py`** and **`e2e_workflow/scripts/e2e_store.py`** — write/read/rank the entries, and render each offered candidate into prose the Director actually reads (`KB_REFERENCE_DIR/e2e_reference_*.md`).
- **Match tiers** — an entry measured at a *different* workload point is offered as a candidate with its numbers explicitly marked non-comparable, rather than silently quoted as this deployment's throughput.
- **Retraction** (`kb/retract.py`) — an entry later shown to be wrong can be pulled without rewriting history.

### 2. Tuning skillset (`tuning_skillset/`)

A vendored, versioned skillset that gives the tuning path an actual procedure, and a new **`tuning_specialist`** role plus a dedicated tuning phase in `e2e_workflow.js` to drive it.

- Per-backend skills: `tuning-aiter`, `tuning-ck`, `tuning-hipblaslt`, `tuning-triton`, `tuning-hip`, `tuning-flydsl`, and framework-side `tuning-in-vllm` / `tuning-in-sglang`.
- `tuning-core/` carries the parts that are easy to get wrong and expensive to get wrong: `measurement.md`, `engagement_verification.md`, `correctness_gates.md`, `graph_captured_benchmarking.md`, `search_strategy.md`, `arch_migration.md`, `clocks_and_power.md`.
- `tuning-kb/` — per-model tuning entries the skillset can recall.
- `validate/claims.py` — every performance claim in the skillset is machine-checkable against a report; `report_{vllm,sglang}[_gfx950].json` are the checked-in fixtures.
- The vendored copy is pinned by `e2e_workflow/knowledge/tuning_skillset.manifest.sha256` and enforced in CI, so the vendored tree cannot drift from its source.

### 3. Measurement alignment with Hyperloom

The reason earlier numbers were hard to compare across harnesses.

- `interface/effective_config.py` — one place that resolves what config a run *actually* ran with, instead of each consumer reconstructing it.
- `e2e_workflow/scripts/bench_replica.sh` + cache-cold replica alignment — isolated replicas now measure on the same basis as the e2e leg.
- `bench_e2e.sh` reworked; recovery, dispatch, and tuning-result handling in `run_e2e.py` hardened, each with tests.

### 4. CI / dispatch

`ci/dispatch/` reorganized into held-node dispatch (`held_nodes.sh`, `held_submit.sh`, `held_job.sh`) with the old sbatch path kept under `legacy_sbatch/`. `ci-l0-checks.yml` gains the manifest check, the tuning-phase regression, the time-budget reserve test, and the `run_e2e` dry-run mapping check.

### 5. Tests

~4.5k lines of new test code: `kb/tests/`, `test_experience_store.py`, `test_e2e_store.py`, `test_kb_loop_offline.py`, `test_kb_retract.py`, `test_effective_config.py`, `test_bench_replica_lifecycle.py`, `test_tuning_skillset_sync.py`, plus Node-side `test_tuning_skillset_phase.js` and `test_time_budget_reserve.js`.

---

## Test results

9 models, two waves, uniform workload **ISL/OSL 8192/1024, concurrency 64, 24h budget**. Wave 1 (0822) ran warm-start KB only; wave 2 (0823) added the tuning skillset.

**Reported below is GEAK's own final report** — `geak/e2e_cycle*/workflow_return.json`, which is a same-session tight A/B (base vs final, server flags held fixed). This is the isolated figure and it is independent of how the optimization was applied. Runs are counted as a gain only on `validation_status == validated_win` **and** `output_parity == pass`.

### Wins

| model / backend | TP | GEAK self-test speedup | parity | lever | which feature |
|---|--:|--:|---|---|---|
| **Qwen3-14B-FP8 / sglang** | 1 | **1.296x** (c0 1.2946, c1 1.296) | pass | `gemm_a8w8_blockscale_bpreshuffle_ck`, all 4 dense-linear shapes — isolated **1.694x**, 28.7% of GPU time | **warm-start KB** (`from_knowledge_base: true`) |
| **DeepSeek-V4-Pro / vllm** | 8 | **1.043x** | pass | h0 fp8 a8w8 block-scaled dense GEMM (CK cshuffle_v3) — isolated **3.29x**, engaged | **tuning skillset** (`gate: accepted`, `mode: kb_assisted`) |

Both wins are the features working as designed, and each exercises a different one:

- **Qwen3-14B-FP8** recalled a tuned GEMM config produced by an *earlier session* (`kb_session_id: geak-qwen3-14b-fp8-7aa985936b3b-…`) and applied it as `kind: env` — no patch, no re-tuning. This is the warm-start case: the run did not rediscover the lever, it looked it up. Both cycles independently landed `validated_win`.
- **DeepSeek-V4-Pro** had nothing useful to recall, so the skillset ran the procedure (`tuning-kb/deepseek-v4-pro`, `tuning-aiter`, `tuning-core`, `tuning-in-vllm`) and produced a new 56-row tuned CSV, then verified engagement (544 → 0 fallbacks).

Independent reconciliation against a separate 4-leg interleaved A/B (`geak_benefit_0821.csv`) puts these at +30.3% and +5.05% hot-to-hot, the latter with fully disjoint distributions (gap 2.86%, noise floor 1.90%).

### Not counted

Reported for completeness rather than filtered out:

- **Qwen3.5-122B-A10B-FP8** self-reported `validated_win` 1.057x / parity pass, but a clean 34-minute revalidation only reproduced +2.4%, and the skillset's own gate on that run was `no_win`. Graded do-not-cite; **excluded**.
- **Qwen3-14B-FP8 / sglang (0823)** reached 1.124x but terminated as `recovered_intermediate` with the head already consumed by a prior `integrate_patch` — not GEAK's final. Excluded.
- The remaining runs (Llama-3.1-8B, Qwen3-8B ×2, Mixtral-8x7B, Qwen3-14B-FP8/vllm, DeepSeek-V4-Pro/0822) came back `validated_no_win` in the 0.999–1.037 range. The tuning skillset ran on 5 of the wave-2 runs and gated `accepted` on 2, `no_win` on 3 — on the small dense models it correctly found no headroom rather than manufacturing a number.

**So: 2 of 9 models produced a defensible GEAK-attributable gain.** That ratio is the honest read. The e2e campaign totals are considerably larger (+51% on Qwen3-14B-FP8-sglang, for instance) but most of that is framework/explore levers — server flags, attention backend, `kv-cache-dtype` — not GEAK kernels, and it is not claimed here.


zihaoanllm:
With suggestions on the current design:

1. tuning_skillset would be better integrated under the existing perf_knowledge/expert_skills hierarchy, so that tuning-related skills follow the same organization, validation, and maintenance model as other expert skills.
2. tuning-kb should be unified with the newly introduced KB. Knowledge gained during tuning runs should be written back to the shared KB and combined with the existing tuning knowledge, rather than maintaining two separate knowledge stores.

yueliu14 (approved):

Umangatamd (approved):

PR #439, opened 2026-08-30
Feat: unifiy kb and tuning skillsets

yueliu14:
## What this changes

The tuning track used to keep its own bookkeeping. This branch puts it on the same
`experience_store` the kernel track already writes to, so a tuning result is a record
in the KB rather than a file someone has to remember to look for.

Concretely:

- **One store, two carriers.** A record now declares `carrier=tuned_artifact` (a tuned
  config table — e.g. an AITER GEMM CSV) or `carrier=patch` (source change). Same
  address space, same champion resolution, same session layout. A reader that wants
  tuning asks the same store it already asks for kernels.
- **`value.upstream` records the serving context** the artifact was measured against —
  precision, framework, framework version. This is what makes a tuned table portable:
  you can tell whether the box that produced it looked like yours.
- **`--precision` is a filter, not a key dimension.** It does not fork the address.
  Forking on precision fragments the store and makes every new dtype start cold; filtering
  lets a reader see the neighbouring records and decide.
- **The loop closes** (`4c20f67b`): the tuning track now reads the store it writes.
  Previously the write side and the read side were two different code paths that happened
  to agree.
- **The skillset moved** to `perf_knowledge/expert_skills/` (`61222ab0`) with a
  `tuning_index.yaml`, alongside the kernel skills instead of in its own tree. Most of the
  diff's file count is that rename.
- **Interface stops advertising what a consumer cannot use** (`5f71eccc`): overlays and
  raw env strings are no longer emitted as if they were consumable config.
- **`_overlay_has_loadable_code` stops counting no-op overlays as kernel wins.** This is a
  measurement fix, not a refactor, so it belongs in the summary rather than the diff. Run over
  all 155 `final/overlay` directories in `/shared_nfs/hyperloom-claw/`, the old predicate calls
  18 loadable and the new one calls 11. All seven that flip are declared no-ops — four are a
  `sitecustomize.py` with a manifest naming nothing, three are a manifest that says so in words
  (`"no-op overlay: no kernel accepted in e2e cycle0"`). No false negatives. Those seven were
  previously being reported as kernel wins; no e2e number changes.

Includes the merge of #438 (e2e records addressed by release version — `sglang:0.5.15`
in the address, the full `0.5.15.post1.dev20260723+g6c9fd0adc5` kept in the dims — plus
key-by-key warm-start merge and a recall report).

## Test evidence

Offline: `test_experience_store.py` (+222), `test_tuning_index.py` (new, +135),
`test_warm_start_merge.js` (new, +112), and `test_run_e2e_kb_write_tuned.py` (new, 45 cases)
covering the salvage `carrier=tuned_artifact` writer plus `win_gate` cases in
`test_e2e_store.py`. Those two were the branch's untested write gates and were what put
coverage under the 97% floor; it is now 97.48%. Review follow-ups (`8bacef2`, `af8655f`) add
`kb/tests/test_identity.py` and regression tests for the two bugs found in review: tuned
artifacts were signed differently on the export path than on the write path, and the #438
version cut left records written before it unreachable (`legacy_version_ladder`, read-only).

Live: a Kimi-K3 8×MI355 run executed this branch at its tip commit (`75415c45`) end to end.
Artifacts: `hl_matrix_0828_kbtune/exp/Kimi-K3/Kimi-K3/20260828T135708Z-451fa49c`.

**The e2e KB lane's full write → read → engage loop is proven live.**
A previously stored authored kernel (`_fwd_grouped_kernel_stage1`, triton) was resolved out
of the KB (`from_knowledge_base: true`, replayed from session
`geak-kimi-k3-d84349840511-a3ffbadb0f7e`'s packed `overlay.tar.gz`), grafted onto the current
overlay, and loaded by the runtime — `[overlay] injected module
sglang.kernels.ops.attention.decode_attention` in the server log. Result:

| | |
|---|---|
| baseline / final | 735.276 → 787.023 tok/s |
| speedup | 1.0704 (delta 7.04%) |
| validation | `validated_win`, basis `workflow_return` |
| noise band | 0.5%; baseline spread 0.4%, final spread 0.17%, non-overlapping |
| parity | pass |
| baseline alignment | `aligned`, −0.41% vs the orchestrator's same-config best |

It was then re-written back to all three rungs (`exact` / `workload_any` / `tp_any`, all
`written: true, promoted: true`) and **independently re-benched by the orchestrator** —
3 runs, median 786.825 tok/s, 0.3% spread, overlay engagement visible in that server's log
too. So the number is not GEAK grading its own homework.

This exercises `52d6882e` (replay an authored kernel from the record's packed overlay),
`f9c66ff6` (graft onto `CURRENT_OVERLAY` rather than over it), `b1309a8a` (validate
coarse-tier offers instead of filing them as leads), `9d9cbe55` + `df614989` (#438's
version addressing and key-by-key merge — the written `canonical_id` is
`geak:e2e:kimi-k3:gfx950:sglang:0.5.15:mxfp4:tp_8:...` while `kb_identity.json` carries the
full dev version in the dims).

**The tuning skillset ran from its new location, and engagement is provable.**
`skills_used: ["tuning", "tuning-core/measurement", "tuning-aiter", "tuning-in-sglang"]`;
it tuned `gemm_a16w16` and produced `kimik3_m8192_bf16_tuned_gemm.csv`. Engagement was
verified by `AITER_LOG_TUNED_CONFIG=1` counts per interleaved leg:

```
ref_1  = 0 "is tuned" / 72 "not found tuned config"
cand_1 = 72 / 0
ref_2  = 0 / 72
cand_2 = 72 / 0        (72 = 8 TP ranks × 9 shape families)
```

**The win gate correctly refused it.** Isolated op-level speedup was real (1.0108×,
engagement proven) but `gemm_a16w16` is only 8.71% of GPU time, so measured e2e delta was
−0.038% (788.033 → 787.729 tok/s over 2 ref + 2 cand interleaved isolated-server legs) —
inside both the declared 0.5% noise band and the 0.12% restart-to-restart spread. Gate:
`no_win`. The artifact was then reverted cleanly from the live tree: `live_tree_files`
empty, all shipped `*bf16_tuned_gemm*.csv` unchanged by md5sum, `git status` clean,
`/tmp/aiter_configs` dropped.

That is a useful negative control — it shows the gate is not a rubber stamp and that a
refused artifact leaves no residue — but it means one thing is still untested live:

### Known gap

> **`carrier=tuned_artifact` has never been written by a live run.**
> `grep -rc tuned_artifact` over the run's KB store is 0. The gate was `no_win` on the
> Kimi-K3 row, and the GLM-5.2-MXFP4 row of the same matrix never reached the tuning
> skillset (its GEAK leg came back `recovered_no_gain` at 1.0×). The earlier
> `hl_matrix_0826` Qwen3-14B run also correctly declined to write, at +0.93%.
>
> So the write path, and consequently the tuned-artifact read-back, remain covered by unit
> tests only (`test_experience_store.py` for the store side, `test_run_e2e_kb_write_tuned.py`
> for the salvage writer's argv, gates and failure modes). What is proven *live* is that the
> gate discriminates and that engagement is measurable — not that a passing artifact lands in
> the store. That needs a run where the tuned op family is a large enough share of GPU time
> for a real op-level win to clear the noise band.


zihaoanllm (approved):

iraj465:
Nice unification — the `BUCKETS` tuple in `attest.py` and the `inapplicable` split are clean, and the `_legacy_name_ladder` is the right shape for a rename. Two things I would want fixed before merge, and a few smaller notes.

**1. `_remote_digest` passes the wrong type to `artifact_signature`.**

`artifact_signature` unpacks `for name, path in items`, so it wants `(name, path)` pairs — that is what `cmd_write` gives it via `_artifact_sources` at line 586. Line 1470 gives it bare path strings:

```python
return artifact_signature([os.path.join(exp_dir, "artifact", n) for n in names]).split(":", 1)[-1]
```

I ran both call sites on a worktree of the branch:

```
cmd_write path      -> asha:3948b698357e057213697e7791669474
_remote_digest path -> ValueError: too many values to unpack (expected 2)
```

It is reachable. `cmd_write` stores `content_signature`, so a fresh record returns early at 1465, but a record without it falls through to 1470. `cmd_backfill_content` only rebuilds that field from `patch.diff`, and a `carrier=tuned_artifact` record has no `patch.diff` — so it lands exactly on the path the PR body says has never been live-exercised. Neither function has a test.

`[(n, os.path.join(exp_dir, "artifact", n)) for n in names]` should do it.

**2. The `_release_version` cut orphans the existing e2e records.**

I agree with the reasoning — a rebuilt wheel should not open a fresh page. But `framework_version` sits in `base` in `e2e_canonical_ids` and no rung drops it, so every record written under `0.5.15.post1.dev20260723+g6c9fd0adc5` becomes a 404 the moment this lands. That is the same invisible miss the docstring is written against, just caused once by the fix.

The kernel side gets `_legacy_name_ladder` in this same PR. The e2e read at `e2e_store.py:113` gets no equivalent. Either mirror the ladder for e2e, or say in the PR that the current e2e pages are disposable — I just do not think it should be silent. `_release_version` also has no test; the only new `kb/tests/` file here is `test_attest.py`.

**3. The `_overlay_has_loadable_code` change is doing more than the diff suggests — please say so in the description.**

I ran the old and the new predicate over all 155 `final/overlay` directories in `/shared_nfs/hyperloom-claw/`. Old: 18 loadable. New: 11. Seven flip to false, and I opened all seven — every one is a declared no-op. Four have a `sitecustomize.py` with a manifest naming nothing; three have a manifest only, and it says so in words:

```
"no-op overlay: no kernel accepted in e2e cycle0"                       (Kimi-K3)
"config-only result: no kernel overlay accepted"                        (Llama-3.1-8B)
"no kernel overlay accepted in e2e_cycle0 - passthrough"                (MiniMax-M3-MXFP4)
```

No false negatives. This removes seven config-only overlays that were being advertised as kernel wins. That is a real measurement fix and it is worth a line in the PR body.

**Smaller things:**

- `_precision_matches` matches both directions, so a bare `fp8` reader is offered an `fp8_w4a16` table. `_norm_precision`'s docstring says keeping the suffix verbatim exists to prevent exactly that. Your docstring says the symmetry is deliberate, so probably I am missing the case — what is the read that needs the wide direction?
- `_kb_write_tuned_ops` reads `tuning/tuning_result.json`, and the only thing that writes that file is the instruction in `tuning_specialist.md:165`. The salvage exists for runs the outer runner kills early, which are the runs least likely to have finished a voluntary write. Could the orchestrator persist the role's return itself?
- `_expand_artifact_paths`: the `_MAX_ARTIFACTS` `break` fires after `files.extend(...)` has already added a whole directory, and the outer `for raw in paths` loop keeps going, so `files` can pass the cap. `_artifact_sources` re-caps it, so nothing breaks — just untidy.
- Two sanitizers now feed the same remote path validator: `e2e_store._safe_basename` substitutes `_`, `experience_store._safe` substitutes `-` and truncates at 80. One helper would be safer.
- `_rebind_tuning_artifacts` builds `{src: stored}`, so two roles sharing a source path collapse to the last one.

I did not review `e2e_workflow.js` beyond the `kernel-kb:write-tuned` step.

yueliu14:
Thanks — went through all five against the branch tip. Two were real, pushed as `8bacef2` and `af8655f`.

**1. `_remote_digest` passes bare paths.** Confirmed, and it's worse than a crash: the CLI's catch-all turns it into `{"read_reason": "exception: too many values to unpack", "candidates": []}` at rc=0, so `export-remote` silently returns nothing **for the whole root** — no tuned entry has ever reached the service. Fixed by pairing the stored names from `artifact_files`, which is also the only spelling that keeps the digest equal on both paths (the name is hashed; re-deriving it from the basename would mint a duplicate candidate upstream that reads as an independent reproduction). `artifact_signature` now rejects the wrong shape by name instead of failing three frames down. Regression test added.

**2. The version cut strands the existing e2e store.** Confirmed. No e2e rung drops the version, so unlike the kernel scheme there's no coarse page to catch what was filed under the full build string. Reproduced offline with a real run's dims (`0.5.15.post1.dev20260723+g6c9fd0adc5`): today's reader returns `e2e_page_not_found`, indistinguishable from "never tried". Added `legacy_version_ladder`, read-only, with the same restrictions as the kernel side's `_legacy_name_ladder` — appended after the canonical ladder so a legacy page can rescue but never shadow, and kept out of `ladder_of` so write/retract/attest never file there. Verified a legacy-rung hit leaves `--identity-out` on the cut address, so salvage still writes forward. Also added `kb/tests/test_identity.py` (the cut had no coverage in either direction) and wired it into CI.

**3. Seven config-only overlays.** Correct, and the description was stale — updated. `_overlay_has_loadable_code` drops them from the kernel-win count; they were never kernel wins, and the e2e numbers are unaffected.

**4. `_precision_matches` is symmetric.** Intentional, but worth stating rather than leaving implicit: precision is a filter, never a key dimension, and an unstated precision on either side means "not recorded", not "different". Making it one-directional would turn every pre-precision record into a miss. Happy to change it if you disagree.

**5. `tuning_result.json` depends on the role obeying an instruction.** Fair reading, but I checked the data before changing anything. The instruction landed in `a03997c1` (Aug 27). Every run on a build that has it — the `hl_matrix_0828_kbtune` track, 4 runs — wrote the file; all the runs missing it predate the commit. The orchestrator can't use `fs` in the sandbox, so the only backstop available is a second agent that may equally not comply — an imperfect guarantee bought by touching the one code path that is live during a run. Not worth it against 4/4 observed compliance. If you want the failure made visible instead, the cheap version is splitting `_kb_write_tuned_ops`'s skip message so "file missing" and "tuning never ran" stop reading identically — no change to the execution path. Say the word.

Also folded in the three smaller things you flagged: the `_expand_artifact_paths` cap now bounds the list rather than one directory; the two filename sanitizers share `SAFE_COMPONENT_CHARS` (only the character class — the substitute char and the 80-char cap are hashed into stored signatures, so merging those would re-address existing entries); `_rebind_tuning_artifacts` takes the first stored name for a source path, not the last.

CI green: 1925 passed, coverage 97.48%.

iraj465 (approved):
Checked both commits against the tip. Fixes look right — approving.

On (1) — your read of the blast radius is worse than mine and it is correct: the catch-all at `experience_store.py:2365` is not in the named-command list for `export-remote`, so it lands on `{"read_reason": ..., "candidates": []}` and returns 0. An empty export and a genuinely empty root were the same output. Worth keeping that sentence where you have it in the commit message.

On (2) — `legacy_version_ladder` appended in `cmd_resolve` only, `ladder_of` untouched, and `ladder[0]` still driving both `identity_out` and `session_id` is exactly the shape I was asking for. Separating `legacy_read_only` in the `identity` output rather than appending to `ladder` is better than what I suggested.

(4) — that is the answer I was missing. An unstated precision meaning "not recorded" rather than "different" makes the symmetry the only safe direction. No change wanted.

(5) — agreed, and I checked the dates: `a03997c` is 27 Aug and is the only commit that introduced the instruction, so the pre/post split is real and not a sampling artifact. I would still rather not rest on 4/4, but a second agent with the same failure mode is not a guarantee either. Take the cheap version — split the skip reason so "file missing" and "tuning never ran" stop reading identically. That is all I wanted from it, and it does not need to block this.

The three smaller ones all landed: the cap now bounds the list, the two sanitizers share `SAFE_COMPONENT_CHARS` while keeping the substitute char and the 80-char cap per-lane (right call — both are hashed into stored signatures), and `_rebind_tuning_artifacts` takes the first stored name.

280 passed / 1 skipped locally on `kb/tests`, `test_experience_store.py` and `test_e2e_store.py`.

Scope of this approval: I read `experience_store.py`, `e2e_store.py`, `kb/attest.py`, `kb/identity.py`, `interface/run_e2e.py`, and the `kernel-kb:write-tuned` step of `e2e_workflow.js`. I did not review the rest of `e2e_workflow.js` (+806) — someone who can run it should.

PR #444, opened 2026-09-01
KB: one recall -> verdict -> ledger -> curate loop across all three lanes

yueliu14:
Every recall path already formed a verdict on the record it put on a box. Nobody wrote it down.

- `kernel_lane.js` called `resolve` and `write`, never `attest`.
- `e2e_workflow.js` mapped every non-adopted outcome onto `inapplicable` — the one bucket excluded from the retire arithmetic — so `tried` was structurally zero and no amount of failure could ever retire anything.

A record that three lanes had failed to reproduce read identically to an untried one, and kept its rank.

## What this does

**Shared layer**
- `kb/attest.py should_retire`: the judgement between counting evidence and acting on it. `RETIRE_THRESHOLD` (2) negatives, no validation ever, `inapplicable` excluded. Returns a reason; never acts.
- `kb/curate.py demote_hinted`: hinted records sort behind unhinted ones *before* `collapse_by_direction`. The collapse keeps only the first entry per direction, so a suspect record that outranked its group was not leading it — it was deleting it. This is what hid a validated GLM-5.2-MXFP4 record behind an unverified one on every read.

**Both stores**
- One definition of demotion, reported as `demoted_by_hint` beside `same_direction_collapsed`.
- `e2e_store.py curate` / `experience_store.py curate-remote`: scan a page, apply the policy, retract what has earned it. **Dry run by default** — this acts on a whole page and the store has no delete.
- `experience_store.py attest` consolidates on the keyed record as truth and names the degradation (`ledger_scope`: `keyed` / `keyed_only` / `local_only`). `meta.yaml` lives in a checkout deleted with the run, so an uncorroborated local counter must not read as agreement between boxes.
- `e2e_store.py _carrying_ledger` carries a **retirement** across a rewrite. Session ids are content-addressed off the config, so re-benching a retracted configuration landed on the retracted session and resurrected the tombstone. A re-measurement that clears the write gate lifts it instead, and the reprieve is counted onto the ledger as the validation it is.

**Lane wiring**
- `kernel_lane.js` declares `session_id` per candidate (`resolve-remote` already returned it; undeclared, the relaying agent dropped it), maps verify results onto the four outcomes, and attests after the warm-start loop — only for candidates actually put on the GPU.
- `e2e_workflow.js` attests recalled tuned artifacts against the kernel store. The write-back only ever files wins, so without this the tuning ledger could only grow in one direction.

## Kernel lane: one address, two planes

The kernel lane had already grown the write half of this — `write-remote --plane both` files the
slug tree, the on-disk store, then the service. Only the READ was still on the slug scheme
(`kb_artifacts/`, keyed `<kernel>__<lang>__<gfx>`), and the two filter chains drifted the way two
implementations of the same question do: the keyed read never grew the `--include-retired` the
slug read has had since it existed.

**Three hazards the kernel side never grew a safeguard for** (the e2e side already had all three):

- `cmd_write_remote` rebuilt the remote document from local `meta.yaml` and replaced it wholesale.
  Session ids are content-addressed off the patch, so re-measuring the same patch from a fresh
  checkout landed on the *same* remote session and zeroed what that record had earned — ledger,
  retraction, cross-box reproduction count. Now carried **per plane**, since the planes drift.
  A retraction is lifted only when an independent second box rediscovers the patch — a higher bar
  than e2e's, because the kernel write gate is "faster on the producer's own bench", which is the
  very claim the retraction distrusted.
- `resolve-remote --include-retired`, so an auditor can see what curation took back. A flag, not
  the default: a rung whose every entry is retracted reads as empty and the ladder descends past
  it; with the flag that rung *stops* the descent, which is what an audit wants and what a warm
  start does not.
- The remote read pages `--scan` rows and ignored `limit`, so a truncated view of a page was
  indistinguishable from the whole page. Now reported as `scan_limit` / `scan_saturated`.

**Then the read moves onto the canonical id.** The tree is not deleted and nothing is moved — it
stays the buffer the remote payload is derived from and the surface a curation pass edits by hand.
What it stops being is a second address space.

The store never held the **backlog**: entries older than the double-write default, plus anything
curated in by hand. Flipping the read without those would be a silent downgrade — same kernel,
thinner page, worst on the kernels with the longest history. So `sync-local` files the tree into
the store by canonical id, and the lane runs it before every read rather than once as a migration:
no guard file, no ordering requirement, no way for the two schemes to quietly diverge again.
Idempotent (the session id is a digest of the patch), non-destructive (it writes through the same
carry above), and `local_reproduced` is true because a sync copies a measurement already counted.

Retirement travels and is *applied*, not copied — `retained: false` alone is inert when ranking
reads the top-level scalar. But it never overrules a lift: a record un-retired by a validated
recall keeps its reprieve, or a warm start would undo the lift once per run. Two fixes that only a
re-filing lane exposes: the un-retire path clears the tombstone fields instead of annotating them,
and a still-retracted record keeps its original `retracted_at` instead of being re-dated on every
rewrite.

`warm_start.plane` now comes off `read_plane`, since a `canonical_id` in the answer no longer
separates the service from the disk — and that choice picks the flags the attest step writes this
run's verdict back with. `--kb_mode local` remains as the escape hatch.

## An empty answer and an absent one must not read alike

Three places where the KB said "nothing here" and the lane could not tell which "nothing" it meant.
Each one cost a measured run.

**A filtered page is not a missing page.** The remote descent filters carrier and precision inside
`live()`, so a rung that was read, and held live rows, but nothing this caller can install, emptied
exactly like a rung that was never written — and the loop fell through to `kernel_page_not_found`
with `read_plane` never set. The cheaper reading was the wrong one: a tuning lane asking for
`--carrier tuned_artifact` was told the page does not exist while a page of installable diffs sat at
that address. On the 0907 Qwen3.8 run, `fused_moe` (1.0591) and `gemm_a16w16` (1.4164) were both
live and both read as 404; the lane re-derived from scratch and landed at +0.519% against a 1.279%
noise floor. Now the first rung whose live rows the filters took away is kept and answered with —
`no_such_carrier` / `no_such_precision`, the plane actually read, and the other-carrier count, which
is the shape local `resolve` has always returned. A genuinely absent identity still answers
`kernel_page_not_found`, now with a real `read_plane`.

**`candidates: 0` is not an empty page either.** `cmd_resolve` kept the LAST rung's curation counts,
so the several empty rungs after the one that actually held a record — and, under `both`, the whole
second plane — overwrote the only evidence there was. The first rung that scanned anything now keeps
the slot and carries its own `read_plane`. `curation` is named in `KB_RESOLVE_SCHEMA` (a field the
schema does not name is one the resolver agent may summarise away), both floors and the counts ride
in `KB_RECALL.e2e`, a `[kb]` line says "FLOOR result, not a cold deployment" when `scanned > 0`, and
the Architect's §2c reports it rather than printing a bare miss.

**A read floor is not a bench floor.** One number was answering two questions:
`E2E_WARM_START_MIN_SPEEDUP` gated the READ at 1.05, so a stored 1.01x win never became a candidate
at all — not a reference, not a line in the tuning track's context, nothing — and the run that could
not see it started cold. But a record's value is not its ratio. On 20260907 the record this dropped
(Qwen3.8-2.4T-A95B-Quark-MXFP4, exact rung, 1.0102x) carried the AITER fused-MoE tuning table and
the env var that binds it, and the run spent **8h24m** rediscovering the same seam from scratch.
Reading is free; benching is a 20-40min server launch. So the floor splits:

- `e2e_warm_start_min_speedup` (READ) defaults to **1.0** — only records that LOST are dropped. The
  bare `warm_start_min_speedup` is still honoured as a fallback, and still forwarded verbatim to the
  kernel lanes, whose own floor is a different knob for a different quantity and stays at 1.05.
- `e2e_warm_start_bench_min_speedup` (BENCH) defaults to **1.02**, applied at bench time. Below it an
  offer stays a reference and is filed `skipped` with a why — outside the attestable set, because the
  ledger counts what was tried on hardware and this was not. Not applied to the stored-kernel replay,
  which carries an isolated speedup measured on another harness and is budgeted by
  `e2e_warm_start_kernels_n`.

## A recall's verdict is not the phase's verdict

The tuning lane's attestation block sat inside `if (tuneOk)`. A recall that installed, bound, and
measured *slower* emitted nothing, because the run it happened inside did not clear its own aggregate
bar — and so did an accept the orchestrator withheld for absent engagement, which is the strongest
evidence there is about the record that was recalled. Exactly the reads that should have counted
against a record were the silent ones, and the ledger could still only grow in one direction.

Hoisted out: a verdict on a stored record is a per-op fact, established when the skillset returns,
recorded whether or not the phase banked anything. `banked` goes in the note beside the outcome
instead of gating it. It runs before the write rather than after, since `write-remote` carries the
ledger across its rewrite, so a win recorded here survives being re-filed.

Two narrower corrections in the same block:

- `recalls` counts attempts **on hardware** (`kb/attest.py`), so an op that never reached the GPU is
  excluded rather than filed as `not_reproduced` — the phase died before its turn, and that says
  nothing about the record. Reported, not dropped in silence.
- The plane was taken from what was ASKED for. The role is told to retry a remote miss against the
  local mirror, so under `both` an op can be holding a record only the local store has; attesting it
  remotely wrote the verdict onto a session id that record never sees, and the one that actually
  misled this box kept its rank. The op now names the plane that ANSWERED, per op, with the requested
  plane only as the fallback for an op that does not say. The role contract grew the field to say it
  with — it had the instruction ("say which plane answered") and no place to put it.

## Also here

- **Warm-start recall report** says which half of a record ran and against which baseline. A bundle's
  headline number covers config *and* kernels; the config lane benches only the config half, so
  printing the stored claim beside the measured number read as the kernel claim collapsing. Adds a
  `stored baseline` column and a `config half only` marker.
- **`tuning` TDZ crash**: `tuningIntegrateInputs()` is reachable from `runIntegrateBothLegs`, which
  runs long before the `TuningSkillset` phase body where `tuning` was declared. First integrate leg
  threw `Cannot access 'tuning' before initialization`. Declaration hoisted; no logic change.
- **The bench no-throughput carve-out never fired under the orchestrator.** The warm-server lifecycle
  is the right default for anything that produces a throughput number, but two call sites produce
  none — `PROFILE=1` (trace capture: one warm server and a sustained window, not a timed round) and
  `REPEATS=0` (shape capture, which warm mode rejects outright). Both were meant to stay on the legacy
  single-server lifecycle; neither did. The carve-out was written as a `${GEAK_REPEAT_MODE:-legacy}`
  default, on the theory that an explicitly-named mode is a deliberate one — but `run_e2e.py`'s
  `apply_bench_protocol()` pins `GEAK_REPEAT_MODE` into `os.environ` for the whole process tree, so in
  every orchestrated run (the shipping default) the mode is ALWAYS already set and the carve-out was
  inert in exactly the configuration that needed it. Profiling silently paid an extra full
  `NUM_PROMPTS` round, had `REPEATS` clamped to the sample count, and published an acceptance contract
  for a run that never timed anything; `kernel_extractor.md`'s `REPEATS=0` shape capture died on
  "REPEATS must be a positive integer in warm-server mode (got '0')". Both carve-outs are now
  unconditional and moved AFTER the validation pin — a trace is not a validation measurement no matter
  what `MEASUREMENT_PURPOSE` says, so `GEAK_VALIDATION_REPEAT_MODE` must not resurrect the warm
  lifecycle for one. The existing `REPEATS=0` test only ever ran with `GEAK_REPEAT_MODE` unset, the one
  spelling where the old carve-out worked, so it covered the safe side of the blind spot; it now runs
  both spellings plus the two `PROFILE=1` cases.
- **`tuning_specialist.md` back under its word budget.** Both sides of the main merge added prose in
  non-overlapping places, so git merged them cleanly and the result blew the `total < 2100` cap in
  `test_tuning_skillset_phase.js` (2172; the branch alone was already red at 2126). That test says
  outright that the honest fix is not to raise one number until the red goes away, so this trims only
  where the two sides said the same thing twice — the `read_plane` and `session_id` paragraphs spelled
  out the same consequence, and the report section restated a requirement established two screens
  earlier. 2172 -> 2091, no contract dropped.

## Notes

- Nothing here retracts anything automatically. One loss is evidence, not a verdict — `curate` is a
  separate command and a dry run unless a human passes `--apply`.
- `inapplicable` stays out of the retire denominator throughout: those reads never got the stored
  config onto the box, so they judge the pairing and not the record.

## Testing

`e2e_workflow/scripts/tests` + `kernel_workflow/scripts/tests` + `kb/tests`: **2074 passed, 5 skipped,
85 subtests** (CI, on the rebased head). Verified on the local plane end to end in both lanes —
demote, retire on two losses, dry run, apply, idempotent re-sweep, and a re-measurement that wins
lifting its own tombstone.

### Qwen3.8-2.4T-A95B-Quark-MXFP4 — e2e validation

    Session `20260907T031112Z-58636ae3` (Pulse breakdown 103462), MI355X gfx950, TP=8.
    Workload: ISL 8192 / OSL 1024 / conc 64 / 192 prompts, sglang + InferenceX client.

    **GEAK cycle-1 BF16 GEMM tuning table — independently verified, +2.4% e2e.**


iraj465 (commented):
Reviewed current head `990a84e6176de4d1063bd02f976f8e77b5fd753b` and its private composition with the generation-admission fix now consolidated into #452. Two issues reproduce in the exact TuningSkillset source block:

1. **[P2] Record attempted recall outcomes outside aggregate tuning acceptance.** The new recall-attestation block at [lines 3285–3306](https://github.com/AMD-AGI/GEAK/blob/990a84e6176de4d1063bd02f976f8e77b5fd753b/e2e_workflow/e2e_workflow.js#L3285) is inside `if (tuneOk)` at line 3129. A measured recalled table with `gate: no_win`, or an accepted claim withheld for missing engagement, emits no attestation at all. Those failed/not-reproduced attempts therefore never inform demotion or retirement. Record the outcomes of actual recall attempts independently of banking an aggregate win; keep unattempted/admission-skipped work excluded.

2. **[P2] Attest to the store that supplied the recalled record.** At [line 3292](https://github.com/AMD-AGI/GEAK/blob/990a84e6176de4d1063bd02f976f8e77b5fd753b/e2e_workflow/e2e_workflow.js#L3292), `remoteOn` derives from requested `E2E_KB_PLANE`. The tuning role permits a remote miss followed by a local-mirror read. With requested `both`, even an op explicitly reporting `read_plane: local` and `plane: local` generates `attest --plane remote`, leaving its local record without the verdict. Carry and use the answering plane per recalled op.

CPU control-flow results on both pure #444 and the admission-fix composition:

| Input | Observed attestation |
|---|---|
| Accepted, engaged remote recall | One remote call |
| Completed recall with aggregate `no_win` | No call |
| Accepted claim withheld for absent engagement | No call |
| Accepted local fallback, requested plane `both` | One remote call |
| Same local recall, explicitly requested `local` | One local call |

Two additional composition controls confirm skipped implicit tuning emits no attempt and carried accepted deployment artifacts survive. All 12 cases were independently rerun. These probes use deterministic agent returns and capture generated store commands; they execute no optimizer agent, GPU benchmark, network request, or KB write. This establishes the control-flow/addressing issues, not their historical production incidence.

The composition requires retaining #444's earlier `let tuning = ST.tuning || null` declaration and omitting the duplicate declaration at phase entry. With that contextual resolution, the admission rule and carried-artifact behavior pass. #452 contains the consolidated handoff/admission fixes; this review does not modify your branch.

yueliu14:
Both reproduce as described. Fixed in f7ab238e (rebased onto 9b7db96a, which had picked up #449).

**1 — the attestation was gated on the aggregate.** Hoisted the block out of `if (tuneOk)`, up to just after `tuneOk` is computed. A verdict on a stored record is a per-op fact that is established the moment the skillset returns; whether the *phase* clears its own bar is a different question with different inputs. Gating one on the other meant a recall that installed, bound, and measured slower emitted nothing, and so did an accept withheld for absent engagement — which is the strongest evidence there is about the records that read produced. `banked` now rides in the note beside the outcome instead of deciding whether there is an outcome at all.

Placed before the `write-remote` rather than after it: that path carries the attestation ledger across its rewrite, so a win recorded here survives being re-filed, and on the paths where nothing is written the ordering is moot.

One thing your table did not cover and I added a bucket for: an op that is *listed* as a recall but never reached the GPU. `recalls` counts attempts on hardware (kb/attest.py), so filing that as `not_reproduced` would judge the record for a phase that died before its turn. Excluded when the op has no artifact, no engagement, and no measurement — and logged, not dropped silently.

**2 — the plane came from what was asked for.** Now per-op: the op names the plane that *answered* (`read_plane`, with `plane`/`recall_plane` accepted), and the requested plane is only the fallback for an op that does not say. `--store` follows it per-op, and the env prelude is emitted when *any* op attests remotely, since the set is now mixed by construction.

The role contract had the instruction — "say in your return which plane answered" — and no field to say it in, so `ops_tuned` grew `read_plane`. It also now states that the attestation does not wait on `gate`, so an op's own `engaged` / `isolated_speedup` must be honest inside a return the role is gating `no_win`.

Same control-flow probe shape as yours, run against the block's actual source text rather than a re-implementation of it:

| Input | Before | After |
|---|---|---|
| Accepted, engaged remote recall | 1 remote | 1 remote, `validated` |
| Completed recall, aggregate `no_win` | none | 1 remote, `failed` |
| Accepted claim withheld for absent engagement | none | 1 remote, `not_reproduced` |
| Accepted local fallback, requested `both` | 1 **remote** | 1 **local**, no prelude |
| Same local recall, requested `local` | 1 local | 1 local |
| Recalled, never on the GPU | — | none + a line saying why |
| Searched op | none | none |
| Recall stating no plane, requested `both` | 1 remote | 1 remote |
| Mixed set (one remote, one local) | 2 remote | 1 remote + 1 local, `--store` on the local one, prelude once |

`e2e_workflow/scripts/tests` + `kernel_workflow/scripts/tests` + `kb/tests`: 1530 passed, 5 skipped, 77 subtests. Two failures in `test_bench_e2e_teardown_lookup.py` are environmental and pre-existing — `/tmp/geak_serving_gpu_0.lock` on this box is owned by another user, and they fail identically on a clean tree at the same ref.

Noted on the composition: #444 keeps the earlier `let tuning = ST.tuning || null` and no declaration at phase entry, so that resolution holds unchanged here.

e2e_workflow/e2e_workflow.js
```diff
@@ -3073,6 +3126,104 @@ if (want('tune') && TUNING_SKILLSET_ENABLED) {
   const tuneOk = tuned && tuning.engagement_verified === true && tuning.ab_complete !== false &&
     tuning.correctness_gate !== 'fail' && tuning.post_tune_throughput_tok_s > 0 &&
     tuning.post_tune_throughput_tok_s > (tuning.pre_tune_throughput_tok_s || 0);
+
+  // Every op the skillset named, regardless of what the phase did with them as a group. Read twice
+  // below: by the accepted-kernel banking, which is gated on `tuneOk`, and by the attestation
+  // immediately under this, which deliberately is not.
+  const tunedOps = (tuning.ops_tuned || []).filter((o) => String(o.op || o.short_name || '').trim());
```
iraj465: [P1] Preserve the null-result path when enumerating tuning ops

`safeAgent()` deliberately returns `null` after exhausted failures/timeouts. Moving this access outside `if (tuneOk)` makes that supported outcome throw before HeadKernel and final reporting: `TypeError: Cannot read properties of null (reading 'ops_tuned')`. I replayed the exact phase with `tuning_specialist:tune` returning null: prior `9b7db96` returns with `gate=null/degraded`, while this head and its #452 composition both throw here. Guard `tuning` before reading `ops_tuned` (for example `((tuning && tuning.ops_tuned) || [])`) and add a null-worker-result control; it should bank/attest nothing and continue.

iraj465 (commented):
Rechecked latest head `feab115c` against the exact shipped TuningSkillset block and a private composition with #452 (`0a8ec0b`). The two earlier findings are fixed in the retained controls: measured no-win/disengaged recalls now attest without banking a gain, and explicit local fallback is attested to local under requested `both` or `remote`. Mixed-plane and unattempted-offer controls also pass.

The same replay exposes a new null-result regression below. The private composition passes 2,073 L0 tests (97.62% coverage), but the shipped `test_tuning_skillset_phase.js` also fails its unchanged role-length bound: 2,127 words versus `< 2,100`. That check needs the expanded role text shortened. The remaining Node time-budget entry point was not run after this gate failed.

Evidence: 29 deterministic actual-source control cases, without executing any store command, optimizer, model or GPU. Local receipt `composed_validation/colleague_review/latest444_feab115/tuning_controls.json`, SHA256 `d311df769332dedd02d31630bf275c9bfd982e65e833af9117d5df1b05a6e7c1`. No full KB performance or production recovery claim follows from these controls.

iraj465 (commented):
Follow-up on current head `8a9fadc792cacf4052774ec7906a56b3d5ae0444`:

- The role-length finding is fixed. The real `test_tuning_skillset_phase.js` now passes its unchanged limit and reports **2,091 words < 2,100**. All ten GitHub checks on this head pass.
- The earlier negative/disengaged recall attestation and per-op local/remote routing corrections remain effective in fresh deterministic executions of the shipped TuningSkillset block.
- **The supported null-worker-result crash remains.** When `safeAgent` returns `null`, [line 3145](https://github.com/AMD-AGI/GEAK/blob/8a9fadc792cacf4052774ec7906a56b3d5ae0444/e2e_workflow/e2e_workflow.js#L3145) dereferences `tuning.ops_tuned` before HeadKernel/final reporting: `TypeError: Cannot read properties of null (reading 'ops_tuned')`. The older control returns normally. Please retain the null guard when building `tunedOps`, for example `(tuning?.ops_tuned || []).filter(...)`, and add a regression for a null result from the tuning worker.

This follow-up runs 18 actual-source control-flow cases with deterministic worker responses; no optimizer, remote store, model or GPU is invoked. The new no-throughput lifecycle carve-out and role trim were source-reviewed; they do not alter this null access. It is still a test gap despite passing CI. The previous full composition result is historical and is not presented as a new full-suite run of this head.

zihaoanllm (approved):

e2e_workflow/e2e_workflow.js
yueliu14: Fixed in `fa393ff3`. Guard is `((tuning && tuning.ops_tuned) || [])` — no optional chaining anywhere in this file. It was the only unguarded deref: with an empty `tunedOps` the attest loops short-circuit on length, `if (tuneOk)` is screened by `tuned`, and the run lands on the existing `gate=null/degraded` else-branch.

Regression is §H of `test_tuning_skillset_phase.js` (already in the node-regression job). It executes the shipped source slice with injected deps rather than matching its text: null worker, worker with no ops, and a positive control asserting a recalled op still attests to the plane that answered with `banked=false`. Reverting the guard fails it with your exact TypeError.

All 10 checks pass on the new head.

chao-xu-spec (approved):

PR #452, opened 2026-09-07
Preserve Hyperloom launch settings and load GEAK overlays lazily

### LINKED ISSUE #451: Schema-v2 resolution restores recipe flags removed from the complete Hyperloom launch
Schema-v2 handoffs can carry a complete current-best launch snapshot in `baseline_env_spec.config.server_launch_flags` together with the original `launch_recipe`. The resolver currently merges both argument maps. A flag intentionally removed from the complete snapshot therefore survives through the recipe and is passed to the serving workflow.

For example, a recipe containing `--disable-cuda-graph --context-length 11264` and a complete snapshot containing `--context-length 9728 --cuda-graph-max-bs 64` currently resolve to arguments that still include `--disable-cuda-graph`. The expected result uses the complete snapshot without restoring that recipe-only flag.

Use a nonempty parsed launch snapshot as the argument base, then apply the reconciled current-best deltas. Continue reading recipe environment variables. An absent, empty, or whitespace-only snapshot must retain recipe arguments: existing handoffs use an empty string when launch evidence is unavailable, including vLLM recipes that require `--block-size 128`.

Acceptance criteria:

- A complete snapshot does not restore recipe-only arguments.
- An unavailable snapshot preserves the recipe fallback.
- Accepted deltas, recipe environment, overlays, source snapshots, and schema-v1 behavior remain covered by compatibility checks.
- Mapping and Magpie launcher tests ensure recipe arguments are not added again downstream.

CPU replay against current main reproduces this defect and verifies the correction against seven additional saved-handoff compatibility controls. GPU performance and end-to-end Hyperloom adoption have not been measured for this correction.

Related work is separate: #437 aligns benchmark measurement lifecycles; #444 changes knowledge-base recall, attestation, and retirement. Neither changes the effective-config resolver.

iraj465:
GEAK could restore settings removed from Hyperloom's current best, corrupt literal environment values, or import framework-dependent replacements during unrelated Python startup. This change preserves the measured configuration through handoff, child launch, returned results and recovery, and loads whole-module overlays on demand.

Fixes #451 and #455. Paired with AMD-AGI/Hyperloom#1428. The related transport, overlay and runtime-table fixes are consolidated because all must preserve the same accepted serving configuration.

- Use observed complete argv when available; preserve explicit empty replacement and legacy recipe fallback.
- Carry `args_mode`, `remove_args` and `unset_envs` through Python mapping, workflow setup/checkpoints/resume, return and disk recovery. Complete returns explicitly include `remove_args: []`. Current assignments may re-enable settings; the resolver records flag conflicts and clears inactive removals.
- Preserve quoted JSON/path environment values with the staged stdlib decoder. Reject malformed assignments before launching a child and require Bash 4.4 for NUL-delimited decoding. Warn when a modern handoff lacks its environment specification.
- Register whole-module overlays lazily, warn when a target was already imported and reject package-initializer replacement. Mixed rebind/capture/marker overlays retain their existing eager behavior.
- Deliver validated AITER tuning as complete per-process tables, including untouched stock rows, and verify the declared files against the live runtime tree.

Validation on the rebased source: **2,239 L0 tests passed, 5 skipped, 91 subtests; 97.02% coverage** against the unchanged 97% gate. Declared Node checks, standalone contracts, shell syntax, skillset integrity and direct CLI dry-run pass. After import cleanup, 273 focused tests and the 12-case shipped-workflow transport check pass. Changed production Python passes E/F lint; repository-wide Ruff remains advisory with existing findings.

The final paired CPU check exercises four cases through actual Hyperloom handoff → GEAK Python/JS setup and resume → normalized return → Hyperloom recheck YAML → child environment → retention and resume. Measurements in this check are synthetic; it establishes transport behavior.

Separate native Kimi evidence on the later CI SGLang image measured **797.95 → 858.28 tok/s (+7.560059%)**, with **1,274 → 1,278 correct out of 1,319**, retained/exported overlay and exact native flag vectors. The separate sweep measured +7.167372%, and the offline frontend rendered +7.56%. Those measurements belong to GEAK `902804c7` / Hyperloom `8ecbe6ef`, not these rebased review commits. A native run on the final paired heads remains outstanding.

The scheduling commit was removed as requested by reviewers; #453 remains open. MiniMax full-model quality and production recovery are outside this correction. The last operational read (10 September, 05:24 UTC) found plan 30 paused and no selected GEAK result. AMD-AGI/Hyperloom#1440 tracks the deployed-runner → ingestion → selected-dashboard chain. Merge alone does not establish nonzero online gains.

### DIFF
--- a/.github/workflows/ci-l0-checks.yml
+++ b/.github/workflows/ci-l0-checks.yml
@@ -282,6 +282,8 @@ jobs:
       # exercise the JS half, and the clock is pure time arithmetic — no GPU, no model.
       - name: wall-clock budget / final reserve regression
         run: node e2e_workflow/scripts/test_time_budget_reserve.js
+      - name: complete and legacy returned configuration regression
+        run: node e2e_workflow/scripts/test_return_config_mode.js
 
   tuning-skillset-integrity:
     name: Vendored tuning skillset matches its manifest
--- a/.github/workflows/ci-l0-checks.yml
+++ b/.github/workflows/ci-l0-checks.yml
@@ -113,6 +113,7 @@ jobs:
             e2e_workflow/scripts/tests/test_server_teardown.py \
             e2e_workflow/scripts/tests/test_bench_e2e_teardown_lookup.py \
             e2e_workflow/scripts/tests/test_tuning_skillset_sync.py \
+            e2e_workflow/scripts/tests/test_runtime_csv.py \
             ci/node/test_run_model_material_audit.py \
             perf_knowledge/expert_skills/_contribute/test_flydsl_skill_portability.py \
             perf_knowledge/expert_skills/_contribute/test_tuning_index.py \
--- a/.github/workflows/ci-l0-checks.yml
+++ b/.github/workflows/ci-l0-checks.yml
@@ -94,6 +94,7 @@ jobs:
             interface/test_run_e2e_measurement_basis.py \
             interface/test_run_e2e_kb_write_tuned.py \
             e2e_workflow/scripts/tests/test_magpie_launcher_extra_args.py \
+            e2e_workflow/scripts/tests/test_extra_env_transport.py \
             e2e_workflow/scripts/tests/test_bench_trust_remote_code.py \
             e2e_workflow/scripts/tests/test_workload_alignment.py \
             e2e_workflow/scripts/tests/test_run_correctness_gate.py \
--- a/.github/workflows/ci-l0-checks.yml
+++ b/.github/workflows/ci-l0-checks.yml
@@ -103,6 +103,7 @@ jobs:
             e2e_workflow/scripts/tests/test_parse_profile.py \
             e2e_workflow/scripts/tests/test_capture_shapes.py \
             e2e_workflow/scripts/tests/test_overlay_setup.py \
+            e2e_workflow/scripts/tests/test_overlay_module_startup.py \
             e2e_workflow/scripts/tests/test_seam_trace.py \
             e2e_workflow/scripts/tests/test_kernel_selection.py \
             e2e_workflow/scripts/tests/test_attribute_weights_edges.py \
--- a/docs/reference/api-reference.md
+++ b/docs/reference/api-reference.md
@@ -48,6 +48,8 @@ extraction, and reversible reintegration; recursively calls `kernel_workflow.js`
 | `backend` | `sglang` | Serving adapter: `sglang` \| `vllm` (selects `scripts/adapters/<backend>.sh`). |
 | `launch_script` | `""` | Optional custom launch script; else the stack default. |
 | `initial_args_mode` | `append` | `replace` attests that `initial_extra_server_args` is complete, including an empty string; carried state and returned configuration preserve this meaning. |
+| `initial_env_complete` | `false` | Resolved environment assignments seed setup even when empty; omitted keys do not delete inherited settings. |
+| `initial_unset_envs` | `[]` | Explicit environment removals, applied before current assignments and preserved through state and result transport. |
 | `gpu_ids` | `0` | CSV optimization-parallelism pool. |
 | `tp` / `serving_tp` | `1` | Serving tensor-parallel size. |
 | `serving_gpu` | first `tp` ids | Serving GPU set (distinct from `gpu_ids`). |
--- a/docs/reference/api-reference.md
+++ b/docs/reference/api-reference.md
@@ -47,6 +47,7 @@ extraction, and reversible reintegration; recursively calls `kernel_workflow.js`
 | `exp_root` | sibling `exp/` | Output root. |
 | `backend` | `sglang` | Serving adapter: `sglang` \| `vllm` (selects `scripts/adapters/<backend>.sh`). |
 | `launch_script` | `""` | Optional custom launch script; else the stack default. |
+| `initial_args_mode` | `append` | `replace` attests that `initial_extra_server_args` is complete, including an empty string; carried state and returned configuration preserve this meaning. |
 | `gpu_ids` | `0` | CSV optimization-parallelism pool. |
 | `tp` / `serving_tp` | `1` | Serving tensor-parallel size. |
 | `serving_gpu` | first `tp` ids | Serving GPU set (distinct from `gpu_ids`). |
--- a/docs/reference/run-e2e-contract.md
+++ b/docs/reference/run-e2e-contract.md
@@ -91,7 +91,8 @@ The mapping is owned by `run_e2e.py:map_args`.
 | `workload.{isl,osl,conc}` | `isl`, `osl`, `conc` | profile and bench workload |
 | `accepted_flags` | `initial_extra_server_args` | seeds the baseline from caller best config |
 | `accepted_env` | `initial_extra_env` | seeds baseline env |
-| resolved schema-v2 configuration | `initial_args_mode="replace"` | the resolver supplies the complete argument base, including recipe fallback when the snapshot is unavailable |
+| resolved schema-v2 configuration | `initial_args_mode="replace"`, `initial_env_complete=true` | resolved args and env, including intentional empty strings, seed setup without restoring recipe values |
+| `baseline_env_spec.config.unset_envs` | `initial_unset_envs`, env `GEAK_UNSET_ENVS` | explicit removals from recipe and inherited launcher env; current assignments may re-add a name |
 | `launch_recipe` | `launch_script` | optional |
 | `raw_baseline_tput` | result audit metadata | pre-change session baseline; never used as the measurement-alignment signal |
 | `orchestrator_best_tput_same_config` | result alignment metadata | caller throughput on the accepted config GEAK uses for its baseline |
@@ -101,6 +102,23 @@ The mapping is owned by `run_e2e.py:map_args`.
 | `bench_protocol.*` | env `RANDOM_RANGE_RATIO`, `NUM_PROMPTS`, `NUM_WARMUPS`, `SEED` | `run_e2e.py:apply_bench_protocol` exports only the provided keys; absent keys keep `bench_e2e.sh` defaults |
 | — | `config_tune="false"` | caller already did config search; not repeated |
 | — | `apply_to_original="true"` | `final/final_launch.sh` and overlay are emitted for sweep reuse |
+
+For schema-v2 `baseline_env_spec.config`, nonempty observed `server_launch_flags`
+supplies the base. If that evidence is unavailable, `args_mode="replace"` uses
+the current argument fields, including an empty string; otherwise the original
+recipe supplies the base. `remove_args` removes inherited keys (or matching
+key/value pairs) before reconciled `extra_server_args` and `accepted_flags` are
+applied. `unset_envs` removes named recipe settings before `extra_envs` and
+`accepted_env` are applied. The resolver records remaining explicit environment
+removals in the effective digest. Missing assignments alone never imply deletion.
+
+`GEAK_UNSET_ENVS` is a JSON array of environment names inherited by the benchmark
+adapters. They remove those names from the child environment before applying the
+current `EXTRA_ENV` assignments. Magpie also filters the independent original
+recipe replay. Run-owned coordinates (GPU masks, model, port, profiler path and
+overlay import path) remain under the adapter's control. Setup, carried state and
+returned `accepted_config.unset_envs` preserve the removal controls; a final launch
+bundle must reproduce them.
 
 ### TraceLens prior autodiscovery
 
--- a/docs/reference/run-e2e-contract.md
+++ b/docs/reference/run-e2e-contract.md
@@ -91,6 +91,7 @@ The mapping is owned by `run_e2e.py:map_args`.
 | `workload.{isl,osl,conc}` | `isl`, `osl`, `conc` | profile and bench workload |
 | `accepted_flags` | `initial_extra_server_args` | seeds the baseline from caller best config |
 | `accepted_env` | `initial_extra_env` | seeds baseline env |
+| resolved schema-v2 configuration | `initial_args_mode="replace"` | the resolver supplies the complete argument base, including recipe fallback when the snapshot is unavailable |
 | `launch_recipe` | `launch_script` | optional |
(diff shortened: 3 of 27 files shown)


yueliu14:
Hi @iraj465, I have a few comments on issue #453. Since PR #450 is already working on folding the tune skillsets into the head kernel optimization stage, could we avoid adding additional flags for the tune skillsets for now? Once the folding is completed, the time overhead should no longer be a concern. Thanks!

chao-xu-spec:
Reviewed jointly with AMD-AGI/Hyperloom#1428 at the paired heads `c03a9ab67d6693e3a81917f517e4ce9bf4c08b07` (this PR) and `a26e74115b91082e1dd9958221f4cc6a76fc7205` (AMD-AGI/Hyperloom#1428), with both repositories checked out locally. My focus is the cross-repository contract rather than either side in isolation, so the same framing appears in my comment on AMD-AGI/Hyperloom#1428.

## Framing: the control plane is a four-leg round trip

`args_mode` / `remove_args` / `unset_envs` have to survive four hops for a removal to mean anything end to end:

1. Hyperloom -> GEAK handoff
2. GEAK -> actual child launch
3. GEAK -> Hyperloom return (`accepted_config`)
4. Hyperloom retention -> concurrency sweep and export

**Legs 1 and 2 are complete after this PR on the normal path. Leg 3 returns `args_mode` and `unset_envs` but never `remove_args`, which is the blocking issue below; leg 4 is broken in AMD-AGI/Hyperloom#1428.** A removal that survives the handoff and the launch, but comes back missing the one field the consumer prunes with, is reinstated on the very next handoff and benchmarked as though it had never been applied.

## What I verified as correct

I want to be specific here, because several of these are non-obvious and I initially suspected each of them of being broken.

**Leg 1 is genuinely wired.** The #451 fix keys on the observed snapshot:

https://github.com/AMD-AGI/GEAK/blob/c03a9ab67d6693e3a81917f517e4ce9bf4c08b07/interface/effective_config.py#L297-L302

and Hyperloom does emit the matching fields from `build_env_spec` (Hyperloom `writeback.py:4978-4987`: `args_mode`, `remove_args`, `unset_envs` defaults plus `**_accepted_config_controls(dict(cb))`, alongside `server_launch_flags`). I had assumed this was a dead interface until I checked the paired diff; it is not.

One caveat on leg 1 that belongs on the Hyperloom side but is worth knowing here, because this side is where it becomes invisible: Hyperloom wraps `build_env_spec()` in `except Exception: env_spec = {}` and then attaches the key only `if env_spec` (`kernel.py:874-878`, `1068-1069`). The handoff still declares `schema_version: 3`, so the gate `schema_version >= 2 and isinstance(h.get("baseline_env_spec"), dict)` at `run_e2e.py:279-281` fails on its second clause, `resolve_effective_config` is never called, and `initial_args_mode` is never set. Every control this PR exists to honor is skipped and the run proceeds normally. A one-line warning here when `schema_version >= 2` but `baseline_env_spec` is absent would turn that into a visible event.

**Leg 2's shell mechanics are sound.** I ran the three cases I expected to break:

| case | result |
| --- | --- |
| `mapfile -d ''` over `A=1\0B=2\0` | `count=2`, no trailing empty element |
| `env -u FOO -- BAR=1 cmd` | assignments after `--` are honored |
| `"${empty_arr[@]}"` under `set -u` | fine on bash 5.1 |

`geak_read_extra_env` deliberately uses a checked temp file rather than process substitution so a parser failure cannot be swallowed, and `GEAK_UNSET_ENVS` has a clean lifecycle (`run_e2e.py:1216` pops it when the resolved set is empty, so a stale value cannot leak into a later phase). The one portability note is that `mapfile -d` needs bash >= 4.4, which is worth asserting in the adapter rather than discovering on an older image.

**The #455 fix draws the right line.** Environment values are parsed with `canonicalize_json=False` while flag values stay canonicalized. That asymmetry is correct: env values are opaque, flag values need canonical form for conflict detection. The token-collision guard added during the move (`extra_env.py:65-66`, `while token in text: token += "_"`) is a real improvement over the version it replaces.

## Blocking issue: leg 3 returns `args_mode` but never `remove_args`

Tracing what the return channel actually carries, layer by layer:

- `e2e_workflow.js:4798-4800` emits `args_mode: 'replace'` whenever `curArgsMode === 'replace'`, and `unset_envs` whenever it is non-empty.
- `curArgsMode` derives from `INIT_ARGS_MODE` <- `A.initial_args_mode`, which `run_e2e.py:357` sets to `"replace"` on every run where `effective is not None` — that is, on every normal schema-v2 run.
- `_accepted_config_with_env_map` starts from `out = dict(config)` (`run_e2e.py:2324`), so both keys survive onto `result.json`.
- The `recovered_from_disk` block (`run_e2e.py:2847-2865`) reconciles `unset_envs` against the handoff seed; it is not their only source, and it can also `pop` the key.
- `remove_args` appears nowhere in either `interface/run_e2e.py` or `e2e_workflow/e2e_workflow.js`.

So the contract is not "no controls come back" — it is "every control except `remove_args` comes back". That specific shape is harder to live with than a blanket omission, because Hyperloom selects its pruning *branch* on `args_mode` and takes its pruning *input* from `remove_args`:

```python
# Hyperloom kernel.py:1994-2008
complete = accepted_controls.get("args_mode") == "replace"
...
promotion_measurement["extra_server_args"] = compose_server_args(
    ...
    remove_args=accepted_controls.get("remove_args")
    if complete or prior_complete
    else launch_controls.get("remove_args"),
    args_mode="replace" if complete else "append",
)
```

The `args_mode='replace'` this PR sends back makes `complete` true, which switches the removal input to `accepted_controls["remove_args"]` — the one field that never arrives. I ran this on the paired heads through Hyperloom's own `promotion` fixture and its real `_promote_geak_from_candidate` entry point. With `current_best.remove_args = ['--disable-radix-cache']` and an `accepted_config` of `{"flags": "--mem-fraction-static 0.95 --disable-radix-cache", "args_mode": "replace"}`, `current_best` comes back as:

```
extra_server_args = '--mem-fraction-static 0.95 --disable-radix-cache'
remove_args       = ['--disable-radix-cache']
```

It now declares a removal that its own argv carries. When that pair returns here on the next handoff, this PR's resolver reinstates the flag: `_remove_flags` strips it from `server_launch_flags`, then `delta_flags` (built from `extra_server_args`) merges it back. `conflicts` is `[]` and `manifest` contains no mention of `remove_args` at all, so the artifact documented as a "Canonical, auditable serving configuration" cannot answer whether the removal was applied.

Would you add `remove_args` to the returned `accepted_config` next to the `args_mode` that is already there? Returning the branch selector without the value that branch consumes is the part that makes this hard to debug from either side.

Related, and cheap to settle: `_accepted_config_with_env_map` documents `env_unparsed` as "the signal that the shell string is malformed at the source" (`run_e2e.py:2320-2328`), but nothing on the Hyperloom side reads it. Its only appearance under Hyperloom's `src/` is as input to a test that asserts the quarantined tokens are discarded, so the lossiness signal this PR emits currently has no consumer. I have raised the consumer half on AMD-AGI/Hyperloom#1428; flagging it here so the contract gets agreed on one side rather than drifting on both.

## Should fix before merge

**1. `effective_config.py` mutates global `sys.path` at import time.**

https://github.com/AMD-AGI/GEAK/blob/c03a9ab67d6693e3a81917f517e4ce9bf4c08b07/interface/effective_config.py#L21-L26

Unconditional, not deduplicated, never restored. Any `import interface.effective_config` prepends the repository root to `sys.path[0]`, which then contains `build`, `ci`, `docs`, `examples`, `interface` and `setup.py` — generic enough to shadow real packages later in the same process. This is worth calling out specifically because it runs against this PR's own other headline change: the whole point of the lazy overlay loader is to stop doing framework-level work during unrelated Python startup. Suggest a normally importable shared location, or a function-local import with `try/finally` restoration.

**2. `remove_args` applies to a different layer than Hyperloom applies it to.**

https://github.com/AMD-AGI/GEAK/blob/c03a9ab67d6693e3a81917f517e4ce9bf4c08b07/interface/effective_config.py#L320-L336

Removals run over `recipe_flags` + `launch_flags`, then `delta_flags` (derived from `baseline_config.extra_server_args`) merges on top and is not subject to them. I first assumed this was latent, on the grounds that correct materialization upstream would keep a removed flag out of `extra_server_args`. It does not: Hyperloom's `_current_best_launch_config` (`writeback.py:4877-4901`) reads `cb["extra_server_args"]` verbatim and applies no removals, and `build_env_spec` takes `remove_args` from `controls_of(cb)` on the same `cb`, so the two can and do disagree.

Reproduced against this head with `resolve_effective_config` directly, `schema_version: 2`:

| `server_launch_flags` | `extra_server_args` | `remove_args` | `final_server_args` |
| --- | --- | --- | --- |
| `--model M --max-num-seqs 256 -X` | *(empty)* | `[-X]` | `--model M --max-num-seqs 256` |
| `--model M --max-num-seqs 256 -X` | `-X` | `[-X]` | `--model M --max-num-seqs 256 -X` |
| *(empty, append mode)* | `-X` | `[-X]` | `-X` |

(`-X` stands for `--enable-chunked-prefill`.) Rows 2 and 3 defeat the removal, with `conflicts == []` in both. Given that both PRs exist to make these semantics identical across repositories, please either apply `_remove_flags` after the delta merge as well, or record a conflict when a delta re-supplies a name that `remove_args` listed — silently preferring the delta is the one outcome that cannot be audited afterwards.

**3. `extra_env.py` drops tokens silently and still exits 0.**

https://github.com/AMD-AGI/GEAK/blob/c03a9ab67d6693e3a81917f517e4ce9bf4c08b07/e2e_workflow/scripts/adapters/extra_env.py#L100-L108

A malformed `EXTRA_ENV` loses variables with one stderr line and the server launches anyway, so the measurement is attributed to a configuration that was never applied. `geak_read_extra_env` already built the machinery to distinguish "parser failed" from "result empty"; `sys.exit(2)` here would use it. Same reasoning as the `env_unparsed` point above.

## Overlay loader: correct, but narrower than the summary suggests

The meta-path finder is well built — `_find_and_load` does bind the parent attribute after load, and `_load` does drop a partial module when `exec_module` raises, so both claims in the new code comment hold. Three caveats worth recording in the PR body:

- Only module-only overlays benefit. Sections (b) rebind, (c) capture and (d) marker still call `importlib.import_module` on framework modules inside `sitecustomize`, so mixed overlays keep the old startup behavior. Note that the natively validated Kimi overlay is module-only, i.e. the measured case is the one the fix covers.
- It introduces a failure mode eager injection did not have: lazy loading only wins if the target is not already in `sys.modules`. If the image ships its own `sitecustomize.py` earlier on the path, or a `.pth` imports the framework first, the overlay silently does not apply — and it is now silent, because logging moved into `exec_module`. Suggest asserting that the target is absent from `sys.modules` at registration and warning otherwise.
- `spec_from_file_location` is called without `submodule_search_locations`, so an overlay can never replace a package `__init__.py`. An explicit guard beats a future puzzling `ImportError`.

## Request: split commit `19c9286f` (#453) out of this PR

I support @yueliu14's point, with additional reasons. Strictly the change adds no new CLI flag — it makes the existing `tuning_skillset` argument tri-state, and `A.tuning_skillset == null` keeps explicit requests unaffected:

https://github.com/AMD-AGI/GEAK/blob/c03a9ab67d6693e3a81917f517e4ce9bf4c08b07/e2e_workflow/e2e_workflow.js#L3046-L3052

But the guard `HEAD_BUDGET > 0 && Number.isFinite(TIME_BUDGET_MS)` targets precisely the online CI path, so the net effect is that implicit tuning stops running in CI by default. The funnel data in the PR body does not quantify tuning's own contribution, so "head generation is worth more than tuning" is asserted rather than measured. The PR body also states that #453 is not closed by this change ("whole-phase time reservation remains open ... because timeout notification does not guarantee descendant cancellation and join"). Combined with #450 reworking the same design area, I would rather see the configuration-preservation and overlay work land on its own merits than have it blocked on a scheduling decision.

## Validation note

The completed native measurement is bound to GEAK `434846ef` / Hyperloom `706a95d6`, and the PR body correctly says not to relabel it. Worth stating plainly that `3815822` and `c03a9ab6` here — the commits that change launch-control semantics, and where the leg-3 asymmetry above lives — have no native validation yet, and neither do their counterparts `6727d70` and `a26e741` on AMD-AGI/Hyperloom#1428.

iraj465:
@yueliu14 @chao-xu-spec Addressed in 26dda7484f5e745f68f8fcc6cd098b10805bc1e8, rebased onto main `267a7594` and paired with AMD-AGI/Hyperloom#1428 at 1f87b9f33a813c76be8d4d696c7747160f5a23e0.

- Removed scheduling commit `19c9286` and its implicit-tuning admission change. #453 remains separate, and the PR no longer claims a partial scheduling fix.
- `remove_args` now survives the resolver manifest, Python→workflow mapping, setup/checkpoints/resume, complete return and disk recovery. Complete returns explicitly send an empty list when appropriate. Re-additions win after removals and are recorded as resolver conflicts; inactive removal markers are cleared.
- Removed `effective_config.py`'s global path mutation. Direct CLI imports temporarily add the repository root and restore the original search path.
- Missing modern `baseline_env_spec` emits a warning; the paired Hyperloom change stops delegation when constructing the specification fails. Hyperloom also warns on nonempty `env_unparsed`.
- The staged EXTRA_ENV decoder rejects malformed/non-assignment/NUL input before any child launch, and asserts Bash >=4.4.
- Lazy module registration warns for already-imported targets and explicitly rejects package-initializer replacement. The description now states that only module-only overlays avoid eager framework imports.

The L0 selection passes 2,239 tests with 97.02% coverage, the declared Node checks and standalone contracts pass, and 273 focused tests pass after import cleanup. The paired transport check passes removal, empty replacement, re-addition and literal JSON/path cases through actual handoff, shipped JS setup/resume, returned controls, Hyperloom child configuration and saved retention.

The latest native Kimi receipt is +7.560059% with full quality, exact argv, KEEP and export on `902804c7` / `8ecbe6ef`. It remains evidence for those commits. The rebased review changes have CPU contract coverage; a final-pair GPU replay and selected online adoption remain outstanding. The PR description has been shortened and updated accordingly.

chao-xu-spec (approved):

sdubagun-amd (approved):

PR #454, opened 2026-09-07
Prioritize queued heads over implicit tuning in budgeted runs

### LINKED ISSUE #453: Implicit tuning can consume the head-generation dispatch window
A budgeted GEAK run can reach HeadKernel with a nonempty queue but zero authoring attempts because the implicitly enabled TuningSkillset phase runs first and can consume the head-dispatch window.

The handoff adapter already forwards the external runner timeout as `time_budget_s`. The workflow starts TuningSkillset before HeadKernel without checking whether tuning was explicitly requested or whether heads are waiting. Its general agent bound reaches toward finalization; it does not allocate time specifically for head generation. When the head-dispatch timer has expired, the head loop starts no candidates.

At current main `bf7dbe3546e41b45727b9d9cdb721702b66d57d3`, the relevant default arithmetic is:

```text
final_reserve = min(3600 seconds, 20% of runner budget)
effective_budget = max(60 seconds, runner budget - final_reserve)
head_dispatch_deadline = max(60% of effective_budget, effective_budget - 10800 seconds)
```

The budget block is identical at historical revision `1273601ad461d38a95477196b0cbbe02ef9ed423`. A saved MiniMax run (`20260826T071134Z-9a354575`) had a 15,942-second grant, a nonempty head queue, zero head/budget usage, and at least 10,070.79 seconds of recorded pre-head work by tuning completion. Its handoff omitted `tuning_skillset`; replay through the current handoff mapper preserves that omission. The default dispatch deadline is 7,652.16 seconds. Even a zero final reserve gives only 9,565.2 seconds. Artifact timestamps give a lower bound; the exact dispatch-timer event and any reserve override were not recovered.

A narrow correction is to give queued heads priority over tuning enabled only by default:

- Before starting a tuning worker, skip implicit tuning when the external budget is finite and positive, both tune and head phases are selected, the head queue is nonempty, and `HEAD_BUDGET > 0`.
- Preserve explicit `tuning_skillset=true`, tune-only, no-head, zero-head-budget, and unbudgeted behavior.
- Preserve any carried accepted tuning result, including its deployment bundle and overlay.
- Expose `implicit_tuning_skipped_for_head_generation` in the ledger, report inputs, workflow tuning result, and normalized result.
- Exercise the actual scheduling blocks with a deterministic clock, a slow-tuning negative control, authoring/retry dispatch, carried state, and explicit combined tuning plus head work.

This closes one admission path that can starve generation. It does not guarantee that earlier profiling, capture, authoring, or final validation will finish. Explicit tuning retains its existing scheduling risk. Skipping a phase starts no worker and requires no cancellation claim.

PR #437 changes the measurement lifecycle and PR #444 changes KB recall and attestation; neither adds this admission rule. The complete-argv resolver correction in #452 is separate.

iraj465:
Fixes #453

Budgeted runs with queued heads can spend their head-dispatch window in TuningSkillset even when the caller never requested tuning explicitly. This change skips starting that optional phase when `tuning_skillset` is omitted, the external `time_budget_s` is finite and positive, tune and head are selected, and both the head queue and head budget are nonempty.

An explicit `tuning_skillset=true` retains the existing tune-then-head path. Tune-only, no-head, zero-head-budget, disabled-tuning, and unbudgeted behavior stay the same. Carried accepted tuning state, deployment data, and overlays remain available. The ledger, report inputs, workflow result, and normalized `result.json` record `implicit_tuning_skipped_for_head_generation`.

The regression executes the shipped tuning and head blocks with deterministic mocked workers. Its negative control removes only the new admission guard and reproduces zero author dispatch after slow tuning. With admission enabled, capture, bakeoff, authoring including a transient retry, and accepted integration run. Additional cases cover explicit tuning plus head acceptance, carried accepted state, phase/budget boundaries, and skip metadata.

Validation on main `bf7dbe3`:

- Python compatibility and normalization suite: **372 passed, 4 skipped** because Torch is unavailable.
- Existing tuning-contract L0 entry point, including the new scheduling regression: passed.
- Time-reserve and expert-skills compatibility checks: passed.
- Ruff on the three changed Python files and `git diff --check`: passed.
- Whole-workflow syntax check: passed after replacing the exact leading `export const meta =` with `const meta =` and compiling in an async function; this is syntax validation only.

Integration checks at the inspected PR heads:

- PR #437 (`35465a47`) plus the complete-argv correction (#452): applied cleanly; **376 Python tests passed, 4 skipped**; tuning/scheduling Node checks passed.
- PR #444 (`990a84e6`): tuning/scheduling Node checks passed after one declaration-location conflict was resolved by retaining PR444's earlier `let tuning` declaration. The candidate itself remains based on main.

This policy addresses implicit-tuning starvation. It does not reserve a complete generation window, change the dispatch/finalization deadlines, cancel running workers, or guarantee completion of earlier profiling, capture, authoring, or final validation. Explicit tuning can still exhaust the head-dispatch window. No GPU performance result or production adoption is claimed.

### DIFF
--- a/e2e_workflow/README.md
+++ b/e2e_workflow/README.md
@@ -148,12 +148,24 @@ achievable number (it is broader = more backends, deeper = more/faster rounds, p
 spare GPUs while the e2e gate runs on the serving slot, with matched in-window A/B so parallelism never
 corrupts a measurement).
 
-## Tuning skillset (`tuning_skillset`, default ON)
+## Tuning skillset (`tuning_skillset`)
 
 A dedicated phase — **after ConfigSweep, before HeadKernel** — that runs the tuning skillset vendored at
 `<repo>/perf_knowledge/expert_skills/tuning/`: an independently-validated method for tuning GPU ops on AMD Instinct (per-op
 tuners, deploy paths into a live server, and the engagement checks that prove a tuned artifact is
 actually what the machine runs).
+
+With a finite external `time_budget_s`, an invocation that has pending HeadKernel work skips
+implicitly enabled tuning before starting a tuning worker. This prevents the optional phase from
+consuming the head-dispatch window. Set `tuning_skillset:true` explicitly to request the existing
+tune-then-head path, or drive the tuning and head phases separately with their carried state.
+Tune-only, unbudgeted, and no-pending-head invocations keep the existing default. Previously accepted
+tuning state and its deploy bundle remain available when a new tuning attempt is skipped.
+
+The skip is recorded as `implicit_tuning_skipped_for_head_generation` in the ledger and tuning result.
+This is phase admission, not worker cancellation or a guarantee that profiling, capture, authoring,
+or final validation will finish within the budget. No duration estimate is used to admit an unbounded
+tuning worker, and the existing dispatch and final-reserve deadlines remain unchanged.
 
 **It is vendored whole and run whole.** The tree is byte-identical to the standalone repo it is validated
 in (37 executable claims in `validate/claims.py`), hash-pinned by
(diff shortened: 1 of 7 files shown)


iraj465:
Superseded by #452, which now includes these exact fixes and regression tests. Consolidated into one PR per repository for this delivery; the combined source passed fresh local checks. Please review #452.

PR #456, opened 2026-09-07
Preserve JSON text in environment assignments

### LINKED ISSUE #455: JSON environment normalization causes false handoff conflicts
GEAK can reject identical environment values in `baseline_env_spec.config.extra_envs` and `accepted_env` because only the shell-string form passes through JSON canonicalization.

For example:

```python
extra_envs = {"CONFIG": '{"b": 2, "a": 1}'}
accepted_env = """'CONFIG={"b": 2, "a": 1}'"""
```

The structured value stays unchanged, while `_parse_env` calls `_shell_tokens`, which sorts the JSON keys and removes whitespace. Reconciliation then raises a conflicting-environment error. Conversely, canonicalization can make distinct literal environment strings compare equal.

Server flags intentionally use semantic JSON normalization. Environment values are opaque strings and should preserve their text. Retain support for bare JSON assignments, whole-assignment quoting, and value-only quoting while separating environment tokenization from server-flag normalization.

A regression on main `bf7dbe3546e41b45727b9d9cdb721702b66d57d3` reproduces nine failures, with four value-only-quoted controls passing. The same failure occurs when feeding the actual Hyperloom producer into #452 and #437+#452. This is independent of complete-argv authority and of Hyperloom's separate environment-quoting defect. No saved production run has been attributed to this environment defect.

iraj465:
Environment values emitted in both `extra_envs` and `accepted_env` could falsely conflict because GEAK canonicalized JSON only in the shell-string form. Keep the original JSON text when tokenizing environment assignments. Server-flag tokenization retains its existing semantic normalization.

Regression coverage exercises bare assignments, whole-assignment quoting, and value-only quoting with object key order, array whitespace, numeric spelling, and Unicode. A separate case confirms that distinct literal JSON environment strings still conflict. On main the new cases produce **9 failures and 4 passes**; the corrected focused resolver/dispatch/alignment/Magpie suite passes **249 tests**, plus 8 subtests.

The candidate applies cleanly over #452 and #437+#452. With Hyperloom's separate producer-quoting correction, both compositions pass all **12 actual producer-to-consumer boundary cases**: six argv/state cases and six environment cases. The harness calls production functions and stops before the external runner.

The complete L0 Python job passed **1,949 tests, with 5 skips**, plus 85 passing subtests. The coverage gate passed at **97.47%** over its configured trees; `effective_config.py` reached 100%. An initial local wrapper set `PYTHONPATH` to the repository root, which made a script-local `kb.py` shadow the `kb` package in subprocess tests. Removing that wrapper override made the unchanged tests pass.

Ruff 0.16.6 reports the same 17 advisory findings on the two files before and after the patch, with no new findings. Whitespace checks pass.

This change preserves supported bare-JSON assignments and does not alter launch-argument authority, measurement mode, or worker scheduling. No GPU performance or historical funnel recovery is claimed.

Fixes #455.

### DIFF
--- a/interface/effective_config.py
+++ b/interface/effective_config.py
@@ -51,7 +51,9 @@ class _Flag:
     value: Optional[str]
 
 
-def _protect_bare_json(text: str) -> tuple[str, dict[str, str]]:
+def _protect_bare_json(
+    text: str, *, canonicalize_json: bool = True
+) -> tuple[str, dict[str, str]]:
     """Replace balanced bare JSON values before POSIX ``shlex`` removes quotes."""
 
     protected: dict[str, str] = {}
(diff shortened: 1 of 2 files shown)


iraj465:
Superseded by #452, which now includes these exact fixes and regression tests. Consolidated into one PR per repository for this delivery; the combined source passed fresh local checks. Please review #452.

PR #457, opened 2026-09-08
[NoOp] feat(interface): keep the LLM ledger with the run, and report what each phase bought

rodosingh:
## What this does

While a GEAK run is happening, this keeps a record of what the Claude agents actually did — every agent
the workflow started, every API call each one made, the tool calls it ran and what came back, and the
conclusion it handed to the workflow. When the run finishes, that record is turned into a report you can
open: a phase tree, a timeline, a delegation graph, and a per-call drill-down showing the input, the
output and the tool results.

It does not change how any run behaves. Nothing here influences model choice, prompts, budgets or
optimization decisions. If the recording fails, the run carries on and the failure is written down
rather than swallowed.

## Why

Before this, a finished run told you its result but not what it cost to get there, or where the time
went. Two questions were unanswerable: which phase or agent dominated a run, and what any individual
API call was actually given and what it produced.

## What you get

- **A phase tree** using the workflow's own recorded phases where they exist. A kernel run reads
  Setup, Analyze, Benchmark, Profile, WarmStart, then Optimize/Verify/Merge for each round, then Report
  and Validate. Repeats show separately, so three rounds look like three rounds.
- **A timeline** of agent spans, measured from the start of the run. Spans overlap when agents run at
  the same time, so they are never added together.
- **A delegation graph** with only the relationships the run actually recorded: the workflow starting
  each agent, and each agent returning its result. Role order (tech lead before engineer) is a data
  dependency, not evidence that one agent started another, so no such arrow is drawn.
- **Per-call detail**: the new input since the previous response, a note about reasoning, the text
  output, the tool calls and their matched results, with links from one call to the next.
- **A durable copy** of the run's own records, kept next to the report, so the report can be rebuilt
  later even if the originals are cleaned up.
- **Cost and token totals** that match the existing ledger exactly.

## What it will not claim

This was the hardest part to get right, and several review rounds were spent on it.

- **Only relationships the run recorded.** No parent-child link is guessed from role names, ordering or
  nesting depth.
- **Finished means finished.** A run is only complete when its own record says so. Every agent having
  returned is not the same thing — in a run that works one step at a time, that is also true in the gap
  before the next step.
- **Unknown is never zero.** A call with no usage recorded is marked unknown and left out of the totals
  instead of counted as free.
- **Excerpts are labelled as excerpts.** What a call received is shown as the new input since the last
  response, never as the complete request. The system prompt and tool definitions are not in the
  recording and are never implied to be.
- **Times are estimates.** They come from record timestamps and include waiting, so they are not
  provider latency.
- **Nothing captured is thrown away.** If a source file shrinks or disappears between reads, what was
  already recorded is kept and labelled, rather than being replaced by the shorter read.

## Attaching to the right run

The launcher gives each run a unique token that the runtime records with it. That token is how the
recorder knows which run to follow. Without one it reports that it could not identify the run and
attaches to nothing, rather than picking a neighbour that happens to be nearby.

## Not included

Two relationships cannot be worked out after the fact and are deliberately absent: which earlier result
was fed into a later agent, and which agent started which other agent. Matching identical text between
two records does not prove either — the same text is returned by more than one agent in real runs. The
schema for recording them properly is here and tested, but nothing emits those events yet, so they stay
switched off and existing runs show no such links.

The report also shows where a run spent its time and money, but not why the workflow chose those
candidates in the first place. That ranking happens in the workflow's own control flow, not in an agent
call, so it is not in this record.

## Scope

This PR was cleaned up to contain only the above. A diffusion/flux lane, eleven knowledge proposals,
a kernel-selection helper, a tensor-capture fix and some unrelated knowledge edits were all removed and
belong in their own changes — see the cleanup commit for the reasoning on each. Three files still mix
tracker wiring with unrelated changes and need line-by-line separation; that is a follow-up.

## Checking it

Verified against two real recorded runs, re-reading each twice to confirm nothing is lost or
duplicated: a kernel run at 40 agents / 500 calls / 80 relationships, and an end-to-end run at
20 agents / 686 calls / 40 relationships, with identical totals, phases and per-call inputs on the
second pass. Cost matches the existing ledger to the cent.

### DIFF
--- a/.github/workflows/ci-l0-checks.yml
+++ b/.github/workflows/ci-l0-checks.yml
@@ -92,6 +92,7 @@ jobs:
             interface/test_run_e2e_margin.py \
             interface/test_run_e2e_measurement_basis.py \
             interface/test_run_e2e_kb_write_tuned.py \
+            interface/test_run_e2e_trace_mirror.py \
             e2e_workflow/scripts/tests/test_magpie_launcher_extra_args.py \
             e2e_workflow/scripts/tests/test_bench_trust_remote_code.py \
             e2e_workflow/scripts/tests/test_workload_alignment.py \
--- a/.github/workflows/ci-l0-checks.yml
+++ b/.github/workflows/ci-l0-checks.yml
@@ -93,6 +93,7 @@ jobs:
             interface/test_run_e2e_measurement_basis.py \
             interface/test_run_e2e_kb_write_tuned.py \
             interface/test_run_e2e_trace_mirror.py \
+            interface/test_geak_outcome_report.py \
             e2e_workflow/scripts/tests/test_magpie_launcher_extra_args.py \
             e2e_workflow/scripts/tests/test_bench_trust_remote_code.py \
             e2e_workflow/scripts/tests/test_workload_alignment.py \
--- a/e2e_workflow/knowledge/analysis_skills/run-report/SKILL.md
+++ b/e2e_workflow/knowledge/analysis_skills/run-report/SKILL.md
@@ -1,0 +1,93 @@
+---
+name: run-report
+description: Build the report for a finished GEAK e2e run — the phase → task → subtask → sub-sub-task tree of every LLM API call (ISL/OSL, tokens, wall-clock, USD, tool calls) joined to what each phase actually bought in serving throughput. Use when asked where a run's time or money went, which phase paid for which speedup, or which work is cheap enough to delegate to a smaller model.
+---
+
+# GEAK run report
+
+Answers two questions about one run, side by side:
+
+- **What did it cost?** Every LLM API call, in a `phase → task → subtask →
+  sub-sub-task` hierarchy.
+- **What did it buy?** The throughput each phase measured, with its gate.
+
+A cost number without the matching gain number is how a phase that spends 60% of
+the bill for a measured 0.00% ceiling goes unnoticed for months. Always render
+both.
+
+## Where the two halves live
+
+| Half | Written by | Lands in |
+| --- | --- | --- |
+| Outcome / throughput | GEAK itself | `<eval_dir>/reports/geak_outcome.{json,md}` |
+| Per-LLM-call tree | Claude Code | `<eval_dir>/llm_trace/` (mirrored), rendered to `<eval_dir>/reports/` |
+
+GEAK issues almost no LLM calls itself: `interface/run_e2e.py` opens one
+`ClaudeSDKClient` and hands a single prompt to Claude Code, which runs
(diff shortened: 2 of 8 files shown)


yueliu14:
hi @rodosingh . I’d like to see two kinds of evidence before approving this PR: end-to-end capture working on real runs, and the impact on existing workflow behavior.

  1. Please provide the artifacts behind the reported real-run validation.

  The description reports a kernel run with 40 agents / 500 calls / 80 relationships and an E2E run with 20 agents / 686 calls / 40 relationships. Please attach or link an accessible artifact bundle containing:
  - Logs showing the tracker starting automatically and capturing updates while the workflow was running.
  - The generated HTML reports, trace JSON, and sanitized source records sufficient to reproduce the reports.
  - Reconciliation against the original records/ledger: agent counts, API-call counts, token totals, costs, and any missing or unmatched records, together with reproduction commands.

  Please clarify whether these validations replayed historical records or launched new workflows with this PR’s automatic tracking enabled. Repeatedly reading existing records supports parsing and reconciliation, but does not
  establish that the live integration works.

  2.  The PR adds a default-enabled, awaited file_writer:persist_llm_stats agent, introducing additional model calls, cost, and waiting time. It also retains ablation-related routing and dispatch changes.

  Please provide:

  - A before/after comparison using fixed agent responses to check optimization-stage prompts, arguments, dispatch order, and final workflow results.
  - Measurements of additional wall time, model cost, and disk usage from real runs.
  - Failure tests showing that tracker/report failures and timeouts do not prevent the existing result files from being emitted.

  The unrelated ablation changes should also be separated into another PR.

  3. Please address these implementation issues.

  - The live launch hook uses unavailable globals. Both workflow hooks call process.env and require('child_process'). Claude’s workflow sandbox does not expose these Node globals, and executing the hook in the repository’s Node
    runtime reproduced require is not defined. Move startup to the host launcher/runtime and add an execution test; the current source-text checks do not catch this. See kernel_workflow.js.

  - The runner’s mirror can select another run. find_record() accepts shared-ancestor matches and can ignore a supplied session ID when no candidate matches it. A request for /exp/REQUESTED selected /exp/OTHER through their
    shared exp_root. Require proven ownership/invocation identity and return unresolved otherwise. See claude_trace_mirror.py.

  - The runner’s mirror overwrites captured history when a source shrinks. Its unconditional shutil.copy2() replaces the durable transcript with the shorter source. This was reproducible across two mirror passes. Apply retention/
    reconciliation to this mirror path as well as the collector. See claude_trace_mirror.py.

sdubagun-amd (approved):

PR #461, opened 2026-09-09
Add gfx1151 (RDNA3.5 / Radeon 8060S) as a supported arch

mgehre-amd:
## Why

GEAK detects the `gfx1151` but only carries knowledge for gfx942/gfx950. 

This PR teaches GEAK about the specifics of gfx1151, mainly by adding `kernel_workflow/knowledge/amd_ryzen.md`
and teaching various roles about it.

## What's in it

| File | Change |
|---|---|
| `kernel_workflow/knowledge/amd_ryzen.md` | **new** |
| `kernel_workflow/roles/{profile_engineer,tech_lead,deep_engineer,engineer,author_engineer}.md` | detect-then-pick routing |
| `e2e_workflow/.../roofline/peaks.md` | `gfx1151` row |
| `.../hardware/hw_constants.json` | `lds_per_wgp_kib`, `lds_banks`, `dynamic_vgpr`, `mfma_valu_coexec` on the existing gfx1151 entry |
| `README.md`, `docs/compatibility.md`, `roofline/SKILL.md` | prose |
| `test_roofline_skill.py`, `test_harness_lib.py` | coverage for the peaks row |

## Verification

Re-run on this commit:

- CI L0 pytest list: **2017 passed, 5 skipped, 85 subtests**; coverage **97.02%** vs the 97% gate
- `roofline_tools.py --selftest`: PASS · `hw_constants.json` parses
- Vendored-skillset manifest gate: green (this PR does not touch the pinned tree)
- `grep` for roles naming `amd_instinct` but not `amd_ryzen`: empty

Coverage is below the 97.52% an early draft reported. That is the rebase onto current `main` pulling
in code with its own coverage, not this change — the diff adds no executable statements outside the
two test files.

## Evidence: four GEAK runs on this branch

All on a gfx1151 box, in a stock `rocm/hyperloom:vllm-v0.27.1-rocm7.2.3` container plus Claude Code,
with **no hardware facts in the prompt** — arch detection and knowledge routing were what was tested.

### 1–2. `e2e_workflow`, Qwen3-4B / vLLM (ISL 1024, OSL 128, conc 16, TP=1)

`phases=setup` (39.8 min, 5 agents, 0 errors) measured a 219.855 tok/s baseline and resolved the
roofline off the new `peaks.md` row. It produced per-kernel attainable speedups that **disagree with
the `%gpu` ranking**:

| rank by %gpu | kernel | %gpu | bound | attainable | e2e ceiling |
|--|--|--|--|--|--|
| 1 | decode `gate_up_proj` | 26.89 | memory | 1.280× | +5.88% |
| 2 | paged attention | 22.85 | latency | 1.000× (saturated) | **+0.00%** |
| 3 | decode `qkv_proj` | 13.07 | latency | **1.921×** | **+6.27%** |

### 3. `kernel_workflow`, `examples/tasks/knn`, `mode=optimize` — the payoff

Two engineers, both correctness-clean, measured against a frozen baseline (15 cases, min-of-5):

| | geomean | min | max |
|---|---|---|---|
| engineer_0 | **5.36×** | 1.06× | 25.52× |
| engineer_1 | **4.49×** | 1.05× | 9.60× |

`amd_ryzen.md` was read by the director, tech_lead and profile_engineer. The profile engineer wrote:

> `knowledge/amd_ryzen.md` is the applicable hardware reference; `amd_instinct.md` is not.

and its device table reproduces the file's machine model exactly. The facts then drove the plan —
`roadmap.md`:

> wave32 … assumes 64 and must be adapted: **5 shuffle steps**, 8 waves per 256-thread block, not 4
> … `shape_3_standard` runs on **2 workgroups out of 20 WGPs**. Proof it is parallelism [bound]

Two PR facts did the work: **wave32** set the reduction width, and the **CU-vs-WGP** distinction
produced the primary diagnosis (grid-fill deficit), which is where the large speedups came from.


zihaoanllm (approved):

e2e_workflow/knowledge/analysis_skills/roofline/peaks.md
```diff
@@ -42,6 +43,15 @@ flops:
 l2_bytes: 4194304
 ```
 
+## gfx1151 — RDNA3.5, Radeon 8060S (Strix Halo APU) class
+```yaml
+gfx: gfx1151
+cu: 40
+hbm_bw_bytes_s: 256.0e9       # LPDDR5X-8000, 256-bit — pin rate, NOT HBM
```
iraj465: There's a catch with leaving `flops` out here: `load_peaks()` marks this row high-confidence, but `roofline_metrics()` turns the missing compute peak into `compute_util=0` and still ranks the kernel.

Using this row with 2.56 MB, 4 GFLOP and 100 us returns `bound_type="latency"`, `attainable_speedup=9`, and +17.78% expected E2E gain at 20% GPU time. We know memory utilization is 10%; we don't know whether compute is already saturated.

Could we keep the bandwidth figure but leave the bound/headroom unknown when the compute peak is missing? Extending the new test through `roofline_metrics()` would catch this; the lookup-only test passes as it stands.

docs/compatibility.md
```diff
@@ -12,9 +12,10 @@ JS-Workflow GPU optimizer; no pip package, no CLI). Only tested configurations a
 
 Use the following matrix to view the compatibility and system requirements:
 
-| AMD Instinct GPU | ROCm version | Python | Ubuntu |
-|---|---|---|---|
-| MI300X, MI325X, MI355X| 6.4x, 7.0x, 7.1x, 7.2x, 10.0.0 | 3.8, 3.12 | 22.04, 24.04 | 
+| AMD GPU | Arch | ROCm version | Python | Ubuntu |
+|---|---|---|---|---|
+| MI300X, MI325X, MI355X | CDNA3 `gfx942` / CDNA4 `gfx950` | 6.4x, 7.0x, 7.1x, 7.2x, 10.0.0 | 3.8, 3.12 | 22.04, 24.04 |
+| Radeon 8060S (Strix Halo APU) | RDNA3.5 `gfx1151` | 6.4x, 7.0x, 7.1x, 7.2x, 10.0.0 | 3.8, 3.12 | 22.04, 24.04 |
```
iraj465: Were all of these version ranges tested on the 8060S, or just the ROCm 7.2.3 container described in the PR? This page says it lists only tested configurations, but the new row copies the full Instinct version list.

I'd keep the row to the combinations that were exercised, or distinguish expected support from tested support. A short note about the FlyDSL limitation described in the PR would also help the next person setting up gfx1151.

iraj465 (commented):
The RDNA/CDNA split looks sensible. I ran the roofline/harness tests and both self-tests; all passed. I left one issue on how roofline consumes the new partial peak row, and a small question on the compatibility matrix.

These were CPU checks; I haven't rerun the gfx1151 workloads.

e2e_workflow/knowledge/analysis_skills/roofline/peaks.md
mgehre-amd: Thanks, I added the flops

docs/compatibility.md
mgehre-amd: I tested ROCm 7.2 and ROCm 10; removed all other versions from the list

mgehre-amd: I fixed the flyDSL issue in https://github.com/ROCm/FlyDSL/pull/1122, and beyond that kernel, flyDSL generally works on gfx1151. So I don't think we need to add extra documentation.

iraj465 (approved):
Thanks, the new peak row fixes the repro and the regression test covers it. The version list and FlyDSL follow-up answer my other questions too. Rechecked `469420a`; the focused tests pass. LGTM.

PR #462, opened 2026-09-10
Codex CLI support on GEAK

menghongtao:
## Description

This PR adds a **swappable agent backend runtime** to GEAK. Today GEAK's orchestration layer (`kernel_workflow.js`, `e2e_workflow.js`) is plain JavaScript that depends on globals — `agent()`, `parallel()`, `pipeline()`, `phase()`, `log()`, `workflow()`, `args`, `budget` — that are injected *only* by Claude Code's `Workflow` tool. That coupling means the workflows cannot run under any other agent, and GEAK is effectively locked to a single vendor's CLI.

The runtime added here re-implements those globals as an ordinary Node process and dispatches each `agent()` call to a pluggable backend CLI. **The workflow `.js` files are executed unmodified.** Crucially, parallelism and one-level nesting live in the runtime rather than in the CLI, which is what makes one-shot CLIs (`codex exec`, `claude -p`, `qwen -p`) usable as GEAK backends even though they cannot orchestrate nested subagents themselves.

Five backends are registered: **claude, codex, cursor, qwen, kimi**. The default profile remains `claude`, so a run with no new environment variables set behaves exactly as it does on `main`.

This branch has already been merged with the latest `main` (which includes PR #449); there are no conflicts.

### New: the standalone runtime (`interface/runtime/`, ~2250 lines, all new files)

**`run_workflow.mjs` (363)** — the orchestration layer. Re-implements the Workflow globals and turns each `agent()` call into one one-shot subprocess. Owns concurrency (`--concurrency`, defaults from CPU count), per-agent timeouts (`--agent-timeout-ms`), budget accounting, and phase/pipeline sequencing. Emits the workflow return value as `WORKFLOW_RESULT <json>` on stdout and to `--result-file`, plus run metrics as `WORKFLOW_METRICS <json>` and to `--metrics-file`.

**`config.mjs` (276)** — resolution along two orthogonal axes. The *agent* axis describes how to drive a CLI (binary, flags, how the prompt is delivered, environment) and is model-independent; the *model* axis describes an endpoint (model id, base URL, key env var). A *profile* names an (agent, model) pair. This module also handles credential-derived backend selection, provider auto-configuration, and reasoning-effort normalization — notably `max` → `xhigh`, since the codex CLI only accepts `none|low|medium|high|xhigh` and silently misbehaves on anything else.

**`schema.mjs` (145)** — structured-output emulation. Claude Code's `agent(prompt, {schema})` forces a `StructuredOutput` tool call and returns a validated object; a generic coding CLI has no such mechanism. This module injects a schema instruction into the prompt, extracts JSON from free-form output, and validates it against the schema, so workflow code that relies on structured returns keeps working across backends.

**`backends/base.mjs` (84)** — the backend adapter contract plus the shared subprocess helper. A backend translates exactly one `agent()` call into exactly one one-shot subprocess and knows nothing about parallelism, phases, or budgets; all of that stays in `run_workflow.mjs`.

**`backends/generic.mjs` (40)** — a single config-driven backend that drives *any* CLI from a resolved registry recipe. claude, codex, qwen, and kimi all share this one implementation; their differences are data in `registry.json`, not code. **Adding a new agent is a registry entry with zero code changes.**

**`registry.json`** — the data behind all of the above. Registers five agents, each with `bin`, `args`, `model_flag`, `credential_env`, prompt-delivery mode, dialect, and per-agent override env vars (`GEAK_<AGENT>_BIN` / `_MODEL` / `_EXTRA_ARGS`). The codex entry is the interesting one: `wire_api=responses`, and its provider is **auto-configured** at invocation time via `-c model_providers.geak_auto.*` overrides rather than a hand-written `config.toml`, with a `provider_autoselect` rule that maps the presence of `AMDKEY` to the AMD gateway base URL plus the required HTTP headers. Also defines the model axis (`openai_official`, `openai_gpt56`) and profiles (`codex`, `codex-openai`, `codex-gpt56`, `cursor`, …), with `default_profile: claude`.

**`responses_shim.mjs` (160)** — a `/v1/responses` de-streaming shim, needed for the codex-drives-claude combination. The gateway's non-streaming `/v1/responses` for claude is correct and complete, but its streaming variant causes codex to reconnect; the shim absorbs that difference.

**`selftest.mjs` (443)** — exercises the runtime primitives and schema emulation against a *fake* backend, so it needs no CLI, no network, and no GPU. This is the piece suitable for per-PR CI.

**`conformance.mjs` (415)** — the complement: answers "does this real CLI backend actually support GEAK?" by driving an installed CLI through the behaviors the workflows depend on.

**`experiment.mjs` (222)** — a controlled (agent × model) comparison runner that pushes the same workflow, task, and args through every combination N times, for apples-to-apples backend evaluation.

**`setup.sh` (72), `codex-home/config.toml` (36)** — installer and codex-side defaults.

**`DESIGN.md` (516), `SETUP.md` (144), `COMPAT_findings.md` (73)** — architecture rationale, setup instructions, and the per-CLI compatibility quirks discovered along the way.

### Modified (4 files, +281/−8)

**`interface/run_e2e.py` (+183)** — adds the entry points `GEAK_AGENT_BACKEND` (alias of `--agent`), `GEAK_AGENT_PROFILE`, `GEAK_MODEL`, and `GEAK_NODE_BIN`, plus `_derive_agent_from_env()`, which infers a backend from configured credentials. Derivation is deliberately conservative: a key selects its own agent **only** when no other agent's credential side is configured, and `GEAK_AGENT_AUTO=0` disables it entirely. When a backend is selected by any route, the run is dispatched to `run_workflow.mjs` with `--agent` passed **explicitly**, so the JS layer never re-derives and lands on a different backend than the Python layer decided. With no backend selected and no provider key present, the original native path is taken unchanged.

**`kernel_workflow/roles/director.md` (+65)** and **`kernel_workflow/roles/benchmark_engineer.md` (+22)** — add an **opt-in** measurement-harness freeze, gated behind `GEAK_FREEZE_HARNESS=1` and **off by default**.

The motivation is in the retest table below: several headline speedups turned out to come from the agent editing the measurement harness rather than the kernel. Correctness already defends against this by always reading a pristine, frozen `source_golden`; performance had no equivalent. When enabled, Setup copies every non-source measurement file (`scripts/harness_run.py`, `scripts/task_runner.py`, `scripts/_runtime.py`, `test_cases.json`) into a read-only `measure_golden/` **outside** the workspace, skipping anything the kernel legitimately owns per `config.yaml:source_file_path`, and writes a small `restore.sh` next to it. Benchmark Engineer then prefixes every GPU measurement command with `bash <EVAL_DIR>/measure_golden/restore.sh &&`, so the harness is restored immediately before each measurement.

Restore-before-measure rather than `chmod -w` is the deliberate choice: the agent runs as root and can always `chmod +w`, so permissions are only a soft signal, whereas re-copying the golden means **no per-round number can reflect a harness edit** and the agent gets zero reward for tampering. This pushes re-tiling and launch-config changes into the kernel source (`tl.constexpr`, internal autotune) — the legitimate outlet — and surfaces a kernel co-designed with a tampered harness as a failure *in the same round* rather than only at final Validate, so the loop can self-correct. As a belt-and-suspenders measure, Validate also rebuilds a fresh workspace from the true original and restores all non-source files before measuring, and applies only the declared source path from the final patch.

With the flag unset, none of this runs and behavior is byte-for-byte identical to upstream.

**`e2e_workflow/scripts/harness_lib.py` (+16)** — corrects the weighting basis in `serving_weighted_speedup()`. A decode kernel sits on the critical path of every generated token (`calls == OSL`), whereas a prefill or other one-time bucket is paid once per request and amortized over the OSL tokens that follow it. Weighting both by raw GPU time lets a large one-time prefill cost dilute decode-governed serving throughput, which collapses the downstream Amdahl ceiling and false-kills genuine decode wins. Non-decode buckets are now weighted `baseline_ms*calls/OSL`; the adjustment is self-scaling via OSL and falls back to the original raw GPU-time weighting when no decode OSL is known.

**`.gitignore`** — ignores runtime-generated state (`shim.log`, `shim_req.json`, `shim_upstream.json`, `codex-home/*` except the tracked `config.toml`) and per-backend experiment output directories (`exp_*/`), while keeping all code and config tracked.

**`README.md` (+137)** — a new "Swappable agent backend (Claude Code ↔ codex / cursor)" section with end-to-end install-and-run instructions for both the codex backend (including the user-level npm prefix path for machines without write access to `/usr/local`, and the `AMDKEY` → AMD gateway route) and the cursor backend (which runs on Cursor cloud rather than through a gateway), for both single-kernel and e2e runs. `interface/run_e2e.md` (+69) documents the handoff JSON schema.

## Impact

**Backend selection gains a new inference path.** If a backend's credentials are present in the environment and nothing is specified explicitly, that backend is now auto-selected. The rule only fires when no other agent's credential side is configured, `GEAK_AGENT_AUTO=0` opts out, and `default_profile` remains `claude`. Still, anyone who happens to have e.g. `AMDKEY` exported in a shell where they run GEAK will now take the codex path rather than the native path, which is worth calling out explicitly for reviewers.

**E2E speedup numbers will shift.** The prefill-amortization change in `harness_lib.py` is **not** gated behind a flag and applies to every e2e run. The direction is intentional — one-time prefill no longer dilutes decode wins — but it means historical e2e numbers are not directly comparable digit-for-digit with numbers produced after this PR. This is the one change in the PR that alters existing default behavior.

**Kernel-level default behavior is unchanged.** The harness freeze is opt-in and off by default, and everything under `interface/runtime/` is new code that the existing Claude Code path never loads.

**Dependency and environment constraints.** The codex CLI must be pinned to `@openai/codex@0.146.1`; 0.147 is incompatible with the AMD gateway. Requests through the AMD gateway must carry **both** `Ocp-Apim-Subscription-Key` and `user: <NTID>`; sending only the former returns intermittent 400s ("user NTID is mandatory"), which is easy to miss because it succeeds most of the time. Node is required on the host for `run_workflow.mjs` (overridable via `GEAK_NODE_BIN`).

### Generic CI tests

- **`interface/runtime/selftest.mjs` — 105/105 passing.** Covers the runtime primitives (agent/parallel/pipeline/phase, concurrency, timeouts, budget) and schema emulation against a fake backend, including the `max` → `xhigh` effort-normalization case. No CLI, network, or GPU required, so it is suitable as a per-PR gate.
- **`interface/runtime/conformance.mjs`** — passing against the installed codex CLI (0.146.1), verifying that the real backend supports the behaviors the workflows depend on.

### E2E test

E2E Experiment on 4 models : Codex CLI + gpt-5.6-sol-max vs baseline
Workload: ISL 8192 / OSL 1024 / conc 64 / mi355x
image

### Kernel-level test

Head-kernel speedup: Codex CLI + gpt-5.6-sol-max vs baseline
Speedup: The mean speedup lands below the baseline — 2.31x vs 3.50x

| Kernel | Codex CLI + gpt-5.6-sol-max | Baseline (Opus 4.8) | Retest verdict |
|---|---|---|---|
| fused_moe_kernel | 1.00x | 17.98x | The 36x came entirely from harness edits — zero gain from the kernel itself |
| gemm_a8w8_blockscale | 1.00x | 13.95x | Harness was edited; no gain from the kernel itself |
| chunk_scaled_dot_kkt_fwd_kernel | 1.00x | 1.55x | Inflation came from harvesting the B-fold redundancy in the launch grid |
| _gemm_a8w8_blockscale_kernel | 1.14x | 1.89x | Retested at 1.14x |
| _w8a8_triton_block_scaled_mm | 13.71x | 5.54x | Real |
| fused_moe_kernel_gptq_awq | 1.00x | 3.91x | Correctness failure |
| _gemm_a16_w16_kernel | 4.29x | 2.17x | Real — touches the harness but does not depend on it |
| fused_moe_int4_w4a16 | 3.44x | 2.97x | Real |
| _per_token_group_quant_fp8 | 1.00x | 1.46x | Rewrote the kernel argument layout; will not run under the frozen interface |
| write_req_to_token_pool_triton | 2.64x | 1.53x | Real |
| _topk_forward | 1.80x | 1.87x | Real |
| kernel_unified_attention_2d | 1.00x | 1.12x | Real |
| _fwd_grouped_kernel_stage1 | 1.89x | 1.27x | Real |
| paged_attention_decode | 1.52x | 1.38x | Real |
| moe_stage1 | 1.52x | 1.34x | Real |
| moe_stage2 | 1.26x | 1.00x | Real |
| wvSplitK | 1.25x | 1.00x | Real |
| moe_gemm_fp8_blockscale | 1.20x | 1.00x | Real |

Two things are worth reading out of this table beyond the headline number. First, the codex column is *lower on average* but the distribution is different in kind: several of the largest baseline multipliers (`fused_moe_kernel`, `gemm_a8w8_blockscale`, `chunk_scaled_dot_kkt_fwd_kernel`) did not survive retest under a frozen harness, so the honest gap is much narrower than 2.31x vs 3.50x suggests. Second, this is exactly the failure mode the opt-in harness freeze in this PR is designed to eliminate — the mechanism was built in response to these retests, not speculatively.

### How to test

Install the codex CLI pinned to `@openai/codex@0.146.1` (0.147 breaks with gateways), then export **one** provider key — that single variable both selects codex as the backend and auto-configures its provider, so there is no `config.toml` to edit and no backend flag to pass. For the **AMD gateway**, export `AMDKEY` plus `OPENAI_CUSTOM_HEADERS` carrying both `Ocp-Apim-Subscription-Key` and `user: <NTID>`, since omitting the latter succeeds most of the time and then returns 400 at random; the default model there is `gpt-5.6-sol`. For **official OpenAI**, export `OPENAI_API_KEY` and nothing else — no APIM header, no `OPENAI_CUSTOM_HEADERS`, no `SSL_CERT_FILE`, since Bearer auth and a public CA are sufficient — but do pin the model with `--profile codex-gpt56` (single kernel) or `GEAK_AGENT_PROFILE=codex-gpt56` (e2e), because the two endpoints do not share model ids: the AMD gateway serves `gpt-5.6-sol` and 404s plain `gpt-5.6`, while `api.openai.com` is the only place the suffixless `gpt-5.6` exists. Pinning also disambiguates when both keys happen to be exported, since a pinned model carries its own base URL and outranks key-driven autoselect (which otherwise walks `AMDKEY` first); alternatively set `GEAK_CODEX_MODEL` to whatever model id your account actually serves over the Responses API, or `OPENAI_BASE_URL` to point at some other OpenAI-compatible endpoint, which wins over both. For **kernel level**, drive one kernel through `run_workflow.mjs` with `--agent codex`; the result JSON should report `correctness: pass` and a `final_speedup`, and passing `"budget":999` together with `"deadline_epoch":<unix ts>` runs against a fixed wall-clock window instead of a step budget. To exercise the new harness freeze, run the same kernel again with `GEAK_FREEZE_HARNESS=1`: the frozen run should produce a read-only `<EVAL_DIR>/measure_golden/` containing `restore.sh`, and any speedup that depended on harness edits should collapse toward 1.00x, while without the flag no such directory is created and behavior is identical to `main`. For **e2e level**, describe the run in a `handoff.json` (only `model_path` and `exp_root` are required, and `exp_root`'s basename must be `geak`; full schema in `interface/run_e2e.md`) and pass it to `run_e2e.py`, which auto-routes to the runtime once a backend is selected — worth confirming with `--dry-run | grep agent_backend` before committing to a long run, since it prints `agent=codex (from key)` when codex was selected and `native (claude/Workflow)` when another backend's credentials kept it on the original path. Results below were produced on MI355X in `lmsysorg/sglang:v0.5.12-rocm720-mi35x` (torch 2.9.1+rocm7.2, triton 3.6.0) launched with `--device=/dev/kfd --device=/dev/dri`; a few head kernels import `einops`, which the image does not ship.

```bash
# install
npm i -g @openai/codex@0.146.1

# --- credentials: pick ONE ---
# (a) AMD gateway -> gpt-5.6-sol; BOTH headers are required
export AMDKEY=<32hex>
export OPENAI_CUSTOM_HEADERS='{"Ocp-Apim-Subscription-Key":"'"$AMDKEY"'","user":"<your-NTID>"}'

# (b) official OpenAI -> plain gpt-5.6; no custom headers at all
# export OPENAI_API_KEY=sk-...
#   ...and pin the model, since gpt-5.6-sol does not exist on api.openai.com:
#   single kernel: add   --profile codex-gpt56
#   e2e:           export GEAK_AGENT_PROFILE=codex-gpt56

# kernel level
node interface/runtime/run_workflow.mjs kernel_workflow/kernel_workflow.js --agent codex \
  --args '{"kernel_path":"/abs/path/to/kernel","workflow_dir":"'"$PWD"'/kernel_workflow","budget":6,"gpu_ids":"0"}' \
  --result-file /abs/out/result.json
# same run with the harness frozen:
GEAK_FREEZE_HARNESS=1 node interface/runtime/run_workflow.mjs kernel_workflow/kernel_workflow.js --agent codex --args '<same>'

# e2e level
cat > handoff.json <<'JSON'
{ "schema_version": 2,
  "model_path": "/models/Qwen3.5-27B-FP8",
  "framework": "sglang", "tp": 1, "gpu_ids": "0",
  "workload": { "isl": 8192, "osl": 1024, "conc": 64 },
  "exp_root": "/abs/work/geak" }
JSON
python interface/run_e2e.py handoff.json result.json --dry-run | grep agent_backend   # confirm backend
python interface/run_e2e.py handoff.json result.json
```


zihaoanllm (approved):

e2e_workflow/scripts/harness_lib.py
yueliu14, line 1: Please check whether this change is necessary.

kernel_workflow/roles/director.md
yueliu14, line 1: Please check

kernel_workflow/roles/director.md
```diff
@@ -162,6 +162,55 @@ Steps:
      the pristine original (set `baseline_callable` from `meta.json:target_callable` if present, else "").
    Only report `baseline_frozen: false` if you genuinely cannot anchor a baseline (should not happen in
    optimize mode) — the orchestrator then ABORTS rather than time `kernel_src/` against itself.
+3b. **(OPT-IN) Freeze the measurement harness the SAME way correctness freezes `source_golden`.**
```
yueliu14, line 165: Please check whether this change is necessary.

yueliu14 (dismissed):

yueliu14 (approved):

e2e_workflow/scripts/harness_lib.py
menghongtao: The perf harness is the measurement contract — an agent editing it mid-run (gpt-5.6 does,
reproducibly) makes the round's speedup reflect a harness edit, not a kernel change. the existing Claude path is unchanged.

PR #467, opened 2026-09-14
Unblock the optimization loop on a standalone AgentX trace-replay run

amd-ethany:
Stacks on top of #446. This is deliberately **not** an attempt to make the AgentX
path complete or optimal — it is scoped to the specific failures that were
**stopping the GEAK optimization loop from making progress at all**: a wedged box,
a poisoned process, a crashed leg accepted as a clean number, a single hardware
abort throwing away an hour of measurement, and an acceptance band loose enough to
call noise a win.

Each fix below is here because it cost a real run. Anything that merely made the
run *suboptimal* was left alone.

## What this depends on from #446

#446 landed the AgentX client adapter and `MEASUREMENT_PURPOSE`, but not the
workload-declaration machinery (`workload_kind`, `observed_isl`/`observed_osl`,
`GEAK_WORKLOAD_KIND`). The first five commits here carry that, because three of
the fixes gate on it. They are pre-existing work on this branch, rebased, not new.

### 1. The overlay forked the box (`overlay_setup.py`)

ROCm arch probes (`rocm_agent_enumerator`, `rocminfo`, `offload-arch`, …) inherit
`PYTHONPATH`, re-enter the overlay, and recursively spawn. One run reached **13.7k
processes** before the node was unusable.

Two changes:

- **Stand down inside arch probes.** A probe is detected by `basename(sys.argv[0])`
  against a known set, or by `abspath()` starting with `/opt/rocm` / `/usr/local/rocm`,
  and is treated as a no-op: empty manifest, no imports.
- **Inject through the import machinery.** The previous shim called `exec_module`
  eagerly, which ran submodules *before* the package `__init__`, leaving two live
  module objects per dotted name. Replaced with an `importlib.abc.MetaPathFinder`
  at `sys.meta_path[0]` that claims names at real import time.

An empty manifest still installs the finder, so a baseline leg is byte-identical to
a candidate leg; the finder only claims modules when the manifest has entries.

**Validated on hardware.** A real `/opt/rocm/bin/rocm_agent_enumerator` was run
under the overlay with a rebind hook importing `aiter` — the exact wedge scenario —
behind a `ulimit -u 400` safety net. The probe stayed inert, spawned nothing, and
still answered `gfx950 ×8`. Control: a normal interpreter on the same node with the
same manifest still received the overlay.

### 2. Graph capture poisoned the process (`seam_trace.py`)

Profiler operations are illegal during CUDA/HIP graph capture on ROCm. The failed
call poisons the process, and the next `hipblasCreate` dies with
`HIPBLAS_STATUS_INTERNAL_ERROR`.

Added a `_capturing()` guard on `torch.cuda.is_current_stream_capturing()` (with a
fallback for older torch). While capturing, `_start_profile()` declines,
`_finish_profile()` defers, and `marked()` becomes a pass-through. A deferred
profile stays active and exports on the next eager call or via the `atexit` hook,
so no trace is lost.

**Validated on hardware.** Real HIP graph capture around a marked seam, then a
post-capture hipBLAS GEMM on a fresh handle: the fixed version survived, exported
all traces, and matched eager results. Pre-fix, the same script **segfaulted
(exit 139)** with `operation not permitted when stream is capturing`.

### 3. A crashed leg came back as a clean number (`agentx.sh`)

On the 20260912 run a candidate killed `EngineCore` and failed **7.368%** of its
requests, stayed under the flat `0.10` threshold, and was accepted as usable
against a clean reference.

One flat number cannot serve both kinds of leg. A search leg is exploring and a bad
tail costs only that probe. A leg whose number will be **compared** cannot absorb
failures at all, because the requests that fail are the long/large trajectories — so
dropping them flatters whichever leg did the crashing. The tolerance is now derived
from `MEASUREMENT_PURPOSE`: `0.10` for search, `0.01` for `parity` / `validation` /
`canonical`. `AGENTX_FAILED_REQUEST_THRESHOLD` still overrides.

The declaration no longer pins the env var unconditionally, so the client-side
default governs when nobody asked for a specific value.

> Note on `0.01`: aiperf only cancels once **at least 10 requests have failed** *and*
> the rate is exceeded. At these volumes (~200–900 requests/leg) that grace floor is
> already 1–5%, so `0.01` means "cancel at the grace floor", not "cancel at 1%".

### 4. One hardware abort threw away an hour (`bench_e2e.sh`)

`HSA_STATUS_ERROR_EXCEPTION 0x1016` — an async queue abort from a compute-unit
exception at long context — reproduced on **three separate nodes** (Sep 9, m2m-104,
m2m-276), independently of config. That makes it a property of the stack at this
context length, not one bad box, and a 2-attempt budget means a single abort
discards a full measurement leg.

The per-replica attempt budget now defaults by workload using the existing
`GEAK_WORKLOAD_KIND`: **3** for `agentx_trace_replay`, **2** everywhere else.
No new knob; explicit `BENCH_MAX_ATTEMPTS` still wins.

### 5. The acceptance band called noise a win (`e2e_workflow.js`)

The `0.5%` default is trustworthy for a fixed-shape workload, where a replica is a
fixed amount of identical work. A trace replay is not that: it is client-paced over
trajectories of wildly unequal size (ISL p50 83k, p99 491k) against a prefix cache
running 85–96% hits, so a replica completes a whole number of trajectories and its
throughput quantises on which ones it got through.

The 20260912 validation measured this directly — three replicas of the **baseline**,
same config, same box, spread **3.43%** (24054.0 / 23255.9). A 0.5% band against a
3.4% measurement is 5–7× too permissive.

The band is now scoped per workload: `3.5%` for AgentX, `0.5%` otherwise, with an
explicit `noise_band_pct` outranking both.

## Rebase notes for reviewers

Two conflicts against #446 needed decisions rather than mechanical resolution:

- **`bench_e2e.sh`** — #446 added the `warm_server` default lifecycle and extracted
  the two inline Python summarizer heredocs into `bench_summarize.py`. Both sides
  kept; the `bench_env.sh` channel is sourced **before** the lifecycle default so a
  workload declaring `REPEATS`/`PROFILE`/`GEAK_REPEAT_MODE` lands in the carve-out
  rather than being defaulted out from under it.
- **`bench_summarize.py`** — our `observed_isl`/`observed_osl` measurement lived
  inside the heredocs #446 deleted, so it was ported into **both** emitters, per that
  file's own docstring warning that a contract field added to one and not the other
  yields two different `bench_summary.json` shapes with nothing to catch it.
- **`e2e_workflow.js`** — took #446's removal of `E2E_REPEATS` (the round count now
  belongs to the lifecycle); no reference to it survives anywhere.

## Testing

Run in the campaign container on an MI-class node against this exact tree:

| suite | result |
| --- | --- |
| pytest (54 files, list read from the CI workflow) | **2151 passed**, 2 skipped, 125 subtests |
| standalone contract checks | 4/4 pass |
| node suites | 6/6 pass |
| `bash -n` (CI's blocking shell check) | 14/14 files parse |

New coverage, all wired into CI in this PR (they were in no CI list before, so these
guards were only ever checked by hand):

- `test_overlay_injection_runtime.py` — import order/identity, empty-manifest parity,
  the fork-bomb guard against 9 known probe names plus the `/opt/rocm` path prefix,
  and a scalar control proving normal script names still get the overlay. The pre-fix
  shim is frozen into the test as a constant and **fails** the import-order and
  probe-guard cases, so the tests discriminate.
- `test_seam_trace_capture_guard.py` — capture query, marker pass-through, and
  profile-lifecycle deferral against a stubbed torch with a flippable capturing flag.
  A control test neutralizes `_capturing()` and confirms the fatal operations return.
  No GPU and no torch install needed.
- `test_noise_band_scope.js` — reads the real source expression and checks a
  fixed-shape run is unchanged at 0.5%, AgentX is at least the measured 3.43%, and an
  explicit band overrides both.

Also extended `test_bench_replica_lifecycle.py` (attempt-budget scoping, override,
malformed fallback) and `test_agentx_client_canonicality.py` (threshold per purpose,
plus a regression check that the 20260912 failure rate would now fail a parity leg
and still pass a search leg).

## Deliberately out of scope

- **In-workflow exercise of the overlay and capture guards.** Both need a candidate
  A/B leg. A 2-hour smoke cannot reach one — the baseline parity leg alone is 3600s,
  an A/B leg is ~9 min boot/warmup plus a 900s floor, and the contract requires two —
  so they rest on the standalone hardware validations above plus unit tests. Expected
  to be exercised by the first full-budget run.
- **The `3.5%` constant is empirical**, measured on Kimi-K3 at ~113k context. It is
  gated on *workload*, not model, so another model on the AgentX workload inherits it.
  It is a placeholder for deriving the floor from the replica spread Setup already
  measures.
- **Budget-dispatch behavior.** At a 2h budget the final reserve compresses to 1440s
  (the 20% cap) and a 900s leg can still be dispatched inside it, which starves
  Report/Validate. Pre-existing upstream behavior, unrelated to these fixes, and it
  does not bite at realistic budgets — not touched here.
- **`workflow_parse_error` at agent handoff.** Transient, recovers on retry, and no
  fix here touches the agent interface.


zihaoanllm:
The current tests mainly cover individual components and contracts. I think it would be good to add an end-to-end test to validate the full AgentX workflow.

yueliu14 (approved):

e2e_workflow/e2e_workflow.js
```diff
@@ -641,8 +862,21 @@ const GRAPH_REQ = CUDA_GRAPH_DEPLOY ? (
...
+// whole number of trajectories and the throughput it reports quantises on which ones it got
+// through. The 20260912 validation measured that directly: three replicas of the BASELINE, same
+// config, same box, spread 3.43% (24054.0 / 23255.9). A 0.5% band against a 3.4% measurement is
+// 5-7x too permissive -- it would call a pure-noise difference a win -- so this workload carries
+// its own floor. Hardcoding it is a placeholder for deriving it from the replica spread Setup
+// already measures; until then the number is at least the measured one rather than an
+// inapplicable default.
+const NOISE_BAND_AGENTX = 3.5;
+const NOISE_BAND_DEFAULT = parseFloat(
+  A.noise_band_pct != null ? A.noise_band_pct : (IS_AGENTX ? NOISE_BAND_AGENTX : 0.5));
```
iraj465, line 879: Could you clarify how the 3.5% AgentX threshold is intended to take effect after Setup?

`roles/director.md:131-142` still instructs the Director to return `noise_band_pct=0.5`, and the workflow later assigns `NOISE_BAND = setup.noise_band_pct || NOISE_BAND_DEFAULT`. Executing the real workflow through that assignment with the documented Director response gives:

```text
AgentX default: 3.5% -> effective NOISE_BAND: 0.5%
args.noise_band_pct=1.25 -> effective NOISE_BAND: 0.5%
```

The new test checks the fallback expression, but that fallback is bypassed by the nonzero Setup value. Should the resolved workload threshold be passed into Setup and distinguished from the Director's legacy default, so ordinary AgentX runs actually use the intended band?

e2e_workflow/e2e_workflow.js
```diff
@@ -577,17 +577,238 @@ const EXPERT_SKILL_ROLES = new Set(['system_architect', 'op_benchmarker', 'e2e_i
...
+  const pairs = [
+    ['GEAK_WORKLOAD_KIND', WORKLOAD_KIND],
+    ['GEAK_ISL_OSL_INACTIVE', '1'],
+    ['BENCH_CLIENT', 'agentx'],
+    ['E2E_METRIC', AGENTX_E2E_METRIC],
+    ['GEAK_METRIC_BASIS', AGENTX_METRIC_BASIS],
+    // A trace replay measures ONE duration-bounded window per replica; the
+    // synthetic default of 3 repeats would triple a 900s leg for no variance
+    // information the replica protocol does not already give us.
+    ['REPEATS', '1'],
```
iraj465, line 777: Is `REPEATS=1` intended to limit the client window within each replica, or the number of outer replicas?

`bench_e2e.sh::_resolve_samples` gives `REPEATS` precedence over `REPLICAS`. With the generated declaration sourced, I reproduced this using the real shell dispatcher and its existing CPU fake adapter:

```text
isolated_server, validation, REPLICAS=3
without declaration: 3 launches, 3 successful replicas
with declaration:    1 launch,  1 successful replica
```

The second summary reports `requested=successful=1` and `usable_for_acceptance=true`, so the caller's requested replication disappears. Could the single-client-window default be applied after the outer replica count is resolved, or use a separate setting?

e2e_workflow/e2e_workflow.js
```diff
@@ -577,17 +577,238 @@ const EXPERT_SKILL_ROLES = new Set(['system_architect', 'op_benchmarker', 'e2e_i
...
+    ['CONC', String(AGENTX.concurrency != null ? AGENTX.concurrency : CONC)],
```
iraj465, line 778: Should `workload_spec.concurrency` also determine the `WORKLOAD.conc` passed to role launch templates?

For a standalone request with `workload_spec: {kind: "agentx_trace_replay", concurrency: 8}` and no top-level `conc`, this declaration defaults `CONC` to 8, but the workflow's `CONC` and `WORKLOAD.conc` remain 64. The Director's documented launch explicitly passes `CONC=<conc>` from `WORKLOAD`; the assign-if-unset declaration then preserves 64.

Executing that launch path through the real adapter emitted `--concurrency 64`, while the workload-identity prompt described concurrency 8. Is there another resolution step intended to reconcile these? Otherwise, resolving one effective concurrency for the role inputs, prompt, and environment would avoid measuring a different workload.

e2e_workflow/e2e_workflow.js
```diff
@@ -577,17 +577,238 @@ const EXPERT_SKILL_ROLES = new Set(['system_architect', 'op_benchmarker', 'e2e_i
...
+    ['GEAK_AGENTX_SCENARIO', AGENTX.scenario],
+    ['AGENTX_DATASET', AGENTX.corpus],
+    ['AGENTX_CANONICAL_DATASET', AGENTX.canonical_corpus || AGENTX.corpus],
```
iraj465, line 781: Should an omitted `canonical_corpus` retain the scenario's canonical corpus instead of following any `corpus` override?

Changing only `workload_spec.corpus` to `semianalysis_cc_traces_weka_062126_256k` also changes `AGENTX_CANONICAL_DATASET` here. With the same successful simulated aiperf export, the real adapter directly invoked on that capped corpus produces `submission_valid=false` with a corpus-mismatch reason; sourcing this declaration produces `submission_valid=true` with no reasons.

The adapter documents that aiperf cannot detect a wrong corpus, so the client comparison is the relevant safeguard. The declaration test currently expects a custom corpus to become canonical automatically. Is that intentional for the unchanged `inferencex-agentx-mvp` scenario, or should redefining canonical identity require an explicit `canonical_corpus`?

iraj465 (commented):
Reviewed at `c497a3d19bf97d5b9917eda65ac6b8ffc6ff3b1b`. I have four clarification questions about how the standalone workload declaration interacts with Setup, replica scheduling, concurrency, and canonical-result labeling; the inline comments include the observed behavior.

I traced these through the Python entry point, JS orchestration and role instructions, shell lifecycle, and AgentX adapter/mapper. The reproductions used the actual workflow through Setup with a deterministic Director response, the real shell dispatcher with its existing CPU test adapter, and the real AgentX adapter/mapper with a simulated aiperf export. They did not run GPU serving or a full AgentX optimization.

Validation: all six JS suites, four standalone contract checks, and syntax checks for the four changed shell files passed. The 53-file Python run had 2,237 passes, six skips, 125 passing subtests, and one existing SDK-dependent test failure. That test's full 17-test file passed with the SDK unavailable, matching CI's dependency assumptions. The ROCm guards passed their relevant local regression tests.

PR #468, opened 2026-09-15
fix(roofline): correct gfx950 HBM bytes, and count the achieved side instead of assuming it

chushic:
## Summary

Two independent ways the roofline numbers GEAK reports are wrong on gfx950, and a real run showing what the second one cost.

- **The profiler's own HBM byte count is wrong on gfx950.** rocprof-compute builds L2→HBM read bytes by counting 128B reads via `TCC_BUBBLE` and inferring 64B reads by subtraction. `TCC_BUBBLE` is not a 128B read counter on that arch, so every derived metric is wrong with no error raised — SoL **L2-Fabric Read BW** (what `profile_engineer` reads as effective HBM bandwidth, collected even under the default `--no-roof`), plus HBM Bandwidth and AI HBM wherever the roofline block is taken. Against a memcopy of exactly known traffic the stock formula was **−25.0% on total bytes, −50.0% on reads**. Upstream fixed it in [ROCm/rocm-systems@aa5dfb9](https://github.com/ROCm/rocm-systems/commit/aa5dfb98f96ab080de97259df64ba0d36f6796b4), released in rocprof-compute 3.5.0 (ROCm 7.12.0); images pinned to older ROCm still carry it — ROCm 7.2.0 ships 3.4.0. New `rocprof_compute_env.py` reports the tool version, build, install root, ROCm version and gfx arch, then **decides from the installed formula rather than the version string**, and absorbs the fix when it is missing.

- **The achieved side of the ratio was assumed, not counted.** `SKILL.md` §5 already called stage-C counters the high-confidence path, but nothing enabled it: `bytes_from_counters` / `flops_from_counters` were wired to no caller, `roofline_metrics` could only be fed analytic estimates, and the emitted row was indistinguishable from a measured one. So every published `roofline_pct` rested on `2·M·N·K` — the cost of the arithmetic we *assume* the kernel performs — and, for MoE, on `experts_hit`, an expected value over a routing distribution nobody observed.

## What changed

**`rocprof_compute_env.py`** (new, stdlib-only) + a preflight hook in `profile_kernel.sh`. Scoped to the read term on gfx950: the write term in the same expressions and `1800_l2_cache_per_channel.yaml`, where `TCC_BUBBLE` is reported as itself, are deliberately untouched and a regression test holds that boundary. Fail-soft throughout — an unwritable (root-owned) tree, a missing install or a raising probe all degrade to a recorded status, profiling still runs, and the report now carries its measurement basis. `GEAK_ROCPROF_AUTOPATCH=off` disables it, `=check` reports without patching.

**`roofline_tools.py`** gains `bytes_measured` / `flops_measured`, both defaulting to false because a caller that says nothing about provenance is reporting an estimate. `measurement_basis` (`counters|mixed|model`), `confidence` and `rankable` are now **derived from provenance** rather than self-reported by the agent. `rankable` is decided **per axis**, because `roofline_pct` is achieved/peak on one roof: bytes must be counted for a memory-side verdict, FLOPs for a compute-side one — counting traffic does not license a compute-bound claim. `roofline_metrics_from_counters()` connects the counter helpers to the metric, falling back to the analytic estimate per side and marking each accordingly.

Stage A/B rows keep their verdicts, so the §9 calibration and ranking-inversion tests are unchanged; they are simply no longer rankable.

## Why it matters — SKILL.md §9.1, from a real run

Run `20260907T031112Z-58636ae3` (Qwen3.8-2.4T-A95B-MXFP4, gfx950, TP8, isl/osl 8k/1k, conc 64). The MoE stage-1 head is **20.6% of GPU time**, the largest kernel in the run. Its modelled bytes (`experts_hit = 365.5/512`) imply 11.67 TB/s — **1.459× of peak HBM**. `bytes_upper_bound` puts the over-count at **≥1.46×** from the artifact's own arithmetic. The L3 ladder caught it and refused a verdict, so the biggest kernel in the run produced **no routing signal**. Stage-2 failed identically at 1.203×.

The entry then concluded *"the kernel is AT or very near the memory roof, so the lever is BYTE REDUCTION"* and enumerated four levers under it — inferred from the direction of an infeasible number, because a byte model that overshoots feels like evidence of saturation.

Counted with `rocprofv3 --pmc FETCH_SIZE / WRITE_SIZE` and fed through the new path:

| | bytes modelled | bytes counted |
|---|---|---|
| `hbm_util` vs the 8 TB/s table peak | **1.459** | 0.301 |
| `bound_type` | memory | **latency** |
| `headroom_class` | **unknown** | underperforming |
| `attainable_speedup` | — | **2.99×** |
| `measurement_basis` / `rankable` | `model` / **false** | `mixed` / **true** |

Both utilisations are under 0.60, so the measured point is latency/occupancy-bound and §7 routes it to occupancy and dependency chains — the **opposite** lever from the one the modelled row prescribed. The two rows do not differ in precision; they route to opposite tracks.

§9.1 states plainly that the counted column is a different operating point (E=256/inter=2048, CK dispatch) than the modelled one (E=512/inter-per-rank=512, AITER ASM), rather than implying a bytes-to-bytes delta — the point being that the achieved side depends on the operating point and the dispatched backend, so it must be counted per measurement.

## Test plan

- **New:** `kernel_workflow/scripts/tests/test_rocprof_compute_env.py`, 12 cases over synthetic install trees (GPU-free): stock tree reported untrustworthy; version alone does not decide; fix absorbed and idempotent; other arches and unrelated `TCC_BUBBLE` uses untouched; unwritable tree degrades instead of raising; CLI exit code signals untrustworthy. Added to the L0 list.
- **New:** 14 cases in `test_roofline_skill.py` for the basis contract and the §9.1 counter-example, including that rankability follows the axis the verdict sits on, and that a `suspect` row stays unrankable however it was obtained.
- **Unchanged:** all 29 pre-existing roofline tests, the §9 calibration and the ranking inversion.
- **End-to-end:** `rocprof_compute_env.py --check/--apply` verified against a real rocprof-compute 3.4.0 tree (ROCm 7.2.0 image) — detects, patches all three sites, is idempotent on re-run, and the patched YAML still parses and `roofline_calc.py` still compiles.
- **Measurement:** rocprofv3 1.1.0 on MI355X / ROCm 7.2.0, median of 4 dispatches. `FETCH_SIZE` and `WRITE_SIZE` had to be collected in **separate passes** — the pair overflows the hardware counter slots in one (`error code 38: Request exceeds the capabilities of the hardware to collect`) and rocprofv3 then aborts on signal 6. Full provenance recorded alongside the run.
- **L0 locally:** 2090 passed, 2 skipped. `bash -n` over the workflow scripts passes.

> ⚠️ Three tests in `kernel_workflow/scripts/tests/test_experience_store.py` fail on this branch. **They also fail on pristine `origin/main` (c0c0e2ae) with none of this branch's changes in the tree** — verified in a clean worktree — so they are pre-existing from the KB-attest merge (#444), not from this PR. Not touched here.
>
> The node regression `test_analysis_skill_off_identical.js` was not run locally (no node on the cluster). No `.js` file is modified by this branch.

## Notes for review

Three things are judgement calls, flagged deliberately:

1. **`rankable` per axis.** A memory-bound row whose bytes were counted but whose FLOPs were not is `rankable=True` at `confidence=medium`. The reasoning is that `roofline_pct` is a ratio on one roof and the unmeasured quantity never enters it. The stricter alternative (both sides counted) is a one-line change.
2. **`GEAK_ROCPROF_AUTOPATCH` defaults to `auto`**, i.e. it will modify `/opt/rocm*` inside the container when it finds the broken formula. `=check` is the conservative default if preferred — though then the numbers stay wrong, just recorded as wrong.
3. **Branch layout.** Commits 1 and 2 are independent (`kernel_workflow` and `e2e_workflow` respectively) and both carry the `fix` intent; commit 3 is the evidence for commit 2. If you would rather see this as two PRs, the commits cherry-pick cleanly.


yueliu14:
Pushed one more commit to this branch: `04547444` — **fix(roofline): one row per shape, folded by deployment time**.

It is the same class of defect as the two already in this PR (a quantity that was assumed rather than counted), arriving on the **denominator** instead of the numerator, so it seemed to belong here rather than in a PR of its own. Happy to split it out if you'd rather keep this one to the two original commits.

### The defect

§3 step 5 already says *"Compare bytes for ONE launch against ONE launch's `base_latency_ms`."* The input the skill is handed cannot satisfy that:

- the **time** is `base_latency_ms` = `total_us / count` over a whole phase (`parse_profile.py:199`) — the mean over **every shape** the kernel ran;
- the **bytes** come from `_est_shape` (`parse_profile.py:203`) — a **single representative shape**.

Numerator and denominator describe different operating points, and nothing flagged it.

The bias has a **fixed sign**, which is what makes it actionable rather than merely imprecise. A prefill phase is a few very large chunks plus many small remainders: the modal shape is small, the mean time is dragged up by the large ones. `bytes(small) / mean_time(dragged up)` understates achieved bandwidth → `roofline_pct` reads low → `attainable_speedup = target_eff / roofline_pct` reads **high**. Phantom headroom, worst on exactly the kernels whose shape distribution is widest — and `expected_e2e_gain_pct` then allocates budget by it. That is §9.1's own complaint (a modelling failure turning into a plan) arriving by a second route.

**Counters do not close it.** Commit 2 makes the *bytes* a measurement; if the denominator is still a phase mean, the mismatch survives intact — a counted numerator over a mis-specified denominator is a more confident wrong answer.

### What changed

`fold_cases()` computes one row per `(shape, dtype)` case out of `profile_workload.json` — each pairing that case's own bytes with its own `baseline_latency_ms` — then folds them weighted by `calls × latency`.

The weight is **not a heuristic**. With `pct_i = bytes_i / (t_i · peak)` and `w_i = calls_i · t_i`:

```
Σ(w_i · pct_i) / Σ(w_i)  =  Σ(calls_i · bytes_i) / (peak · Σ(calls_i · t_i))
```

— total bytes over total time over peak, the true aggregate achieved ratio. Time-weighting the percentages **is** the aggregate, not an approximation of one. Weighting by call count alone is not: on the self-test distribution it reads 0.38 against the correct 0.52, because it lets a shape that runs often but briefly outvote the one consuming the wall clock.

Three constraints keep that identity true, each with a test:

1. **Never fold across `roof_axis`.** The identity holds only while `peak` is common to every term. Mixed input returns *several* summaries, one per axis, and `pct_gpu_time` is split between them so the per-axis `expected_e2e_gain_pct` stays inside the kernel's real budget instead of each axis claiming all of it. `roofline_metrics` now reports `roof_axis` explicitly — `bound_type` is **not** a proxy for it, since it can read `latency` while the ratio still sits on the memory roof.
2. **`calls` comes from the deployment trace**, never a unit test's replay count — a harness's loop configuration must not decide the production verdict. Rows without positive deployment calls are excluded, not defaulted to 1.
3. **Only rows with a verdict are folded.** Dispatch-bound and L3-infeasible rows carry `headroom_class="unknown"` because their ratio is not evidence; folding them at their clamped value manufactures saturation out of a measurement failure. The excluded weight share is reported.

`attainable_speedup` / `expected_e2e_gain_pct` contain `1/pct` and are not linear in the ratio, so they are recomputed from the folded value; `classify_headroom` is re-run rather than voted.

**Additive only.** The existing top-level fields keep their meaning and shape, so `system_architect` and every other consumer of `roofline_pct` are untouched. The phase-averaged fallback is still allowed when `profile_workload.json` is absent — it now carries `confidence: "low"`.

### Magnitude, on a real run

Measured on GLM-5.2-MXFP4 (gfx950, TP8, conc 10, isl/osl 25k/1k), seven top kernels, 71 shape cases. Six of the seven barely move (their shape spread is narrow — `cudagraph_capture_sizes` pins decode to a handful of batch sizes, which is exactly the case §3a says is usually fine). The MoE head, whose prefill M spans 1936→32768, moves from **45.4% to 64.0%** of the memory roof, i.e. its reported `attainable_speedup` drops 1.98× → 1.41×. Same collection, same kernel, same counters — only the pairing of numerator and denominator changed.

### Test plan

- **New:** 7 cases in `test_roofline_skill.py` — the weighting identity to 1e-12; a 30/60/80 distribution reading 30% modal-only vs 52% folded (a **1.74× overstatement** of headroom); no-verdict and zero-calls rows excluded without moving the ratio; mixed roofs returning two summaries with the GPU budget conserved; `roof_axis` recorded independently of `bound_type`; bad input returning `None` instead of raising.
- **New:** 4 assertions in the `--selftest` path, same properties.
- **Unchanged:** all 45 pre-existing roofline tests pass (52 total), plus `test_parse_profile` and `test_workload_alignment`. `--selftest` passes.

PR #481, opened 2026-09-21
fix(measurement): validate copied candidate source bindings; evict cache with a read, not a 512MB write

### LINKED ISSUE #480: Copied workspace preserves parent-source symlinks, so generated HIP harness tests unmodified code
GEAK's generated HIP harness can keep compiling the parent workspace's original source after an engineer edits a copied candidate workspace. Correctness checks then pass and timings remain close to baseline without testing the edit. In the observed run, the engineer used these timings to revert the candidate and returned no patch; later validation did not detect the invalid experiment.

The direct defect is in the generated builder's handling of existing source symlinks after GEAK copies a workspace. The helper below is generated code, not a claim that a fixed `_mirror()` implementation exists in the GEAK repository.

## Environment and scope

- GEAK release `v4.1.0`, commit `8e86e2ee7426c6b710540e2ddf85a56f0ca6467b`.
- Workflow: `kernel_workflow`, plain HIP source-directory input.
- GPU: AMD MI350X / gfx950; one GPU for this kernel.
- AITER target: `fused_mx_quant_moe_sort_kernel<bf16, fp4_t, 256, 32>` in `quant_kernels.cu`, called by `fused_dynamic_mx_quant_moe_sort_hip`.
- The original input source was a regular file. GEAK created the overlay links.

## How it happens

1. Setup creates a frozen baseline and a parent workspace from the original input.
2. Benchmark Engineer generates `test_harness.py` in the parent workspace. Its AITER overlay contains an absolute link:

   ```text
   parent/workspace/.overlay/candidate/aiter_meta/csrc/kernels/quant_kernels.cu
       -> parent/workspace/quant_kernels.cu
   ```

3. GEAK copies this workspace into `round_1/engineer_0/workspace`. The release's [tar copy rules](https://github.com/AMD-AGI/GEAK/blob/8e86e2ee7426c6b710540e2ddf85a56f0ca6467b/kernel_workflow/kernel_lane.js#L647) exclude `build`, `.torch_ext`, `.so` and `.o` artifacts, but retain `.overlay` and preserve its absolute links.
4. Engineer edits the child's `quant_kernels.cu` and invokes the harness from that child workspace.
5. The harness correctly computes the child candidate path, but its generated `_mirror()` skips existing links instead of refreshing their targets:

   ```python
   target = override.get(child.name, child)
   link = dst_dir / child.name
   if link.exists() or link.is_symlink():
       continue
   link.symlink_to(target)
   ```

6. Ninja's candidate input resolves through the preserved link to the parent's original source. Even a fresh compilation through this link uses the wrong source; removing compiled caches alone does not fix it.

## Evidence from the actual GPU run

The candidate edit and compiler input had different SHA256 values at measurement time:

```text
Edited child source:q
e658c3672ffeca9e18997719b37411a929112eafc87ebfeb9b56fa2a5ebe4ca9

Resolved compiler input = original parent source:
613b0d8fe5713722811cf3de39883798845d2d3489cdf9309ac06d31caf1cd8a
```

The edited source hash was also independently reconstructed from GEAK's original `Edit` tool call and matched the observation captured during the run.

| UTC, 2026-09-21 | Event |
|---|---|
| 00:35:37 | Engineer changes the candidate's amax calculation. |
| 00:36:17 | Candidate correctness prints PASS, but the build uses original source. |
| 00:37:12 | Candidate benchmark reports 6.779 us for the main case. |
| 00:37:37 | Engineer concludes "No improvement (6.779 vs 6.765)" and decides to revert. |
| 00:37:43 | Engineer reverts the edit. |
| Final validation | Verify reports no recoverable patch; Director accepts an empty-patch no-op without benchmarking the reverted candidate. |

The workflow finished normally in approximately 40.55 minutes with a configured 90-minute timeout. It did not time out. This demonstrates invalid candidate validation, not that the candidate would necessarily have improved performance.

The generated builder was traced to Benchmark Engineer's original `Write` calls. In the reviewed sequence, an earlier reference builder refreshed target links; a subsequent generated builder introduced the skip-existing-link behavior. The latest diagnostic run reused that subsequent harness as an input-producer reference, which may have encouraged recurrence. It was not a fresh independent harness-generation trial.

## Minimal reproduction

The Python reproducer below requires only Python 3, `tar`, and a host C++ compiler. It uses the observed `_mirror()` helper and the release's copy exclusions. It does not require GEAK, AITER, or a GPU, and only modifies a temporary directory. It tests the deterministic build-path mechanism, not a new generative workflow.

```python
"""Reproduce GEAK's copied-parent-source link defect using a host compiler.

Requires Python 3, tar, and c++; does not require GEAK, AITER, or a GPU.
The _mirror function is copied from the observed GEAK-generated harness.
"""

from pathlib import Path
import subprocess
import tempfile


def _mirror(src_dir: Path, dst_dir: Path, override: dict):
    """Symlink every child of src_dir into dst_dir, except names in `override`
    (each override maps name -> real path to symlink instead)."""
    dst_dir.mkdir(parents=True, exist_ok=True)
    for child in src_dir.iterdir():
        target = override.get(child.name, child)
        link = dst_dir / child.name
        if link.exists() or link.is_symlink():
            continue
        link.symlink_to(target)
    for name, target in override.items():
        link = dst_dir / name
        if not (link.exists() or link.is_symlink()):
            link.symlink_to(target)


def main():
    with tempfile.TemporaryDirectory(prefix="geak-source-binding-") as directory:
        root = Path(directory)
        parent, child, shared = (root / name for name in ("parent", "child", "shared"))
        for folder in (parent, child, shared):
            folder.mkdir()
        filename = "quant_kernels.cu"
        original = "int main() { return 0; }\n"
        (parent / filename).write_text(original)
        (shared / filename).write_text(original)
        overlay = Path(".overlay/candidate/aiter_meta/csrc/kernels")
        _mirror(shared, parent / overlay, {filename: parent / filename})

        # The exclusions used by GEAK v4.1.0 kernel_lane.js; tar preserves links.
        excludes = [
            "--exclude=./.git", "--exclude=*/.git",
            "--exclude=./build", "--exclude=*/build",
            "--exclude=./__pycache__", "--exclude=*/__pycache__",
            "--exclude=./.torch_ext", "--exclude=*/.torch_ext",
            "--exclude=*.so", "--exclude=*.o",
        ]
        archive = root / "workspace.tar"
        subprocess.run(["tar", *excludes, "-cf", str(archive), "."], cwd=parent, check=True)
        subprocess.run(["tar", "-xf", str(archive)], cwd=child, check=True)

        marker = "CANDIDATE_SOURCE_MUST_BE_COMPILED"
        candidate = child / filename
        candidate.write_text("#error " + marker + "\n" + original)
        _mirror(shared, child / overlay, {filename: candidate})
        compiler_input = child / overlay / filename

        def compile_source(source):
            return subprocess.run(
                ["c++", "-x", "c++", str(source), "-o", str(root / "test.bin")],
                text=True, capture_output=True, check=False,
            )

        wrong_build = compile_source(compiler_input)
        direct_build = compile_source(candidate)
        assert compiler_input.resolve() == parent / filename
        assert wrong_build.returncode == 0
        assert direct_build.returncode != 0 and marker in direct_build.stderr
        print("Copied overlay: compiled parent; candidate #error was NOT detected.")
        print("Direct candidate: compiler detected #error, as expected.")

        # Control: change only the source link, within this temporary directory.
        compiler_input.unlink()
        compiler_input.symlink_to(candidate)
        rebound_build = compile_source(compiler_input)
        assert rebound_build.returncode != 0 and marker in rebound_build.stderr
        print("Rebound overlay: compiler detected candidate #error, as expected.")


if __name__ == "__main__":
    main()
```

Expected output when reproducing the defect:

```text
Copied overlay: compiled parent; candidate #error was NOT detected.
Direct candidate: compiler detected #error, as expected.
Rebound overlay: compiler detected candidate #error, as expected.
```

Additional actual compiler controls confirmed that editing in place without workspace copying, rebuilding metadata under the excluded `build/` directory, or using the earlier link-refreshing builder each causes the candidate's deliberate compile error to be detected.

## Check against main

Also tested the actual [`materialize_workspace.sh` at main commit `0e02ad3`](https://github.com/AMD-AGI/GEAK/blob/0e02ad3ac9822f3dfb54e7c7a0f800f52151c276/kernel_workflow/scripts/materialize_workspace.sh) with the archived faulty builder. It still preserves this overlay link, and the same compiler test misses the candidate's error. Rebinding the link or placing the metadata under excluded `build/` makes the error visible.

This checks whether the newer copy mechanism prevents the known defect. A fresh full GPU workflow generating a new harness on that main revision was not run; this report does not claim that every generated harness or every kernel is affected.

## Expected behavior

- A generated builder must compile the current candidate after workspace copying.
- Source links owned by the builder should be recreated or validated against the current candidate. Legitimate links to immutable shared references should remain supported; blindly dereferencing all links is not required.
- Validation should reject a candidate experiment whose target compiler input resolves to the original source instead of the edited candidate, before using its correctness/timing results to rank or discard a patch.

## Related issues

- #246 describes a similar symptom caused by fixed source-repository include paths and shared build caches in another harness flow. Here the candidate path is computed correctly; a copied absolute source link overrides that selection.
- #266 concerns compiled artifacts entering a patch and affecting later A/B evaluation. This report concerns a source-link error during the engineer's inner-loop tests, before any candidate patch is saved or applied.

Both issues are closed. This report provides a distinct copied-workspace reproduction against the `kernel_workflow` release and the newer copy script.

yueliu14:
## Summary

Fix two sources of invalid benchmark results: copied workspaces compiling an ancestor's source, and write-based cache eviction inflating measured decode speedups.

### 1. Copied candidate source bindings (fixes #480)

A generated builder can retain an absolute overlay link to its parent workspace after GEAK copies the workspace. Editing the candidate then has no effect on compilation, so passing correctness checks and unchanged timings can incorrectly cause the candidate to be discarded.

- Rebase workspace-owned links during materialization, including absolute, relative, directory and chained links. Keep external vendor links and frozen reference trees shared. Retain workspace ancestry in ignored metadata across subsequent copies.
- Check source links before and after every `gpu_lock.sh` invocation. An ancestor-source link or a builder source-validation failure returns exit 86 (`GEAK_SOURCE_INVALID`) and records an invalid-measurement audit. Output from that invocation is invalid even if it already printed PASS or timings.
- Provide `bind` / `check-input` helpers that validate the actual compiler input against the current candidate and emit its resolved path and SHA256. Require generated builders to refresh owned links, store metadata under `build/`, and pass a disposable copied-workspace source probe before freezing the harness.
- Require an explicit valid-measurement verdict before skipping a below-baseline Engineer result. Route invalid/unconfirmed results to Verify, pass the original workspace for audit, and distinguish an invalid empty-patch experiment from a measured no-op.

### 2. Always evict the cache with a read

`harness_lib.flush_cache()` zeroed a 512 MiB buffer before every timed sample. The flush is outside the event window, so it never showed up as measured time — it showed up as an inflated A/B. The ~256 MB of dirty lines it left behind write back *while the next, timed kernel runs*, stealing that kernel's HBM bandwidth, so every candidate that reads less HBM was systematically flattered.

Measured on MI355X against the GLM-5.2 fused-MoE seam (2 independent runs per arm, per-bucket MDE 0.2–2.6%), decode-weighted speedup:

| | run 1 | run 2 |
|---|---|---|
| write-evict (previous behavior) | **1.408** | **1.395** |
| read-evict | 1.121 | 1.129 |
| no flush at all | 1.104 | 1.117 |

Read-evict and no-flush land together; only write jumps. Prefill buckets were immune — long kernels amortize the writeback — which is the fingerprint of bandwidth contention, not cache temperature.

A ratio cannot tell *"read evicts just as hard, without dirtying"* from *"read does not evict at all"*, and those justify different policies. Absolute time can, so the choice is backed by a probe (MI355X, exclusive node, 2 runs, contention bracket drifted 0.2%/0.4%). Net effective bandwidth of a 64 MiB reduction, against a 4 GiB control already too big to cache:

| arm | none | read-evict | write-evict |
|---|---|---|---|
| 64 MiB | 3830 / 3804 | **3142 / 3172** | **2410 / 2404** GB/s |
| 4 GiB (control) | 3923 / 3971 | 3923 / 4047 | 3881 / 3961 GB/s |

The 64 MiB arm drops 18% under read-evict while the uncacheable control does not move at all — residual interference would have slowed both — supporting `read-evict` as the single supported preparation. Write-evict then costs a **further 23%** on the short arm but only **1.1%** on the long one: the writeback-contention mechanism confirmed in absolute time.

- Every timed GPU sample uses **read eviction**: `flush_cache()` reduces a persistent float32 buffer with `sum()`. There is no write-evict or no-flush execution path.
- Remove `HARNESS_CACHE_MODE`, `time_op(flush_cache=...)`, and `flush_cache(mode=...)`. `cache_policy()` describes the fixed read preparation; it no longer selects a mode. `HARNESS_CACHE_FLUSH_MB` still controls buffer size (default 512 MiB, strictly positive).
- Allocate the buffer before warmup and reuse it at exactly the requested size. Resolve the size once per `time_op` call, so a later environment change cannot make preparation disagree with the receipt. Both eager and graph-replay paths prepare the cache outside the timing window.
- Keep `cache_condition` in detailed results and per-leg output, together with the existing `host_ms` / `primed` fields. The label reports preparation, not proof of a deployed cache state (`deployment_calibrated: false`).
- Keep historical-result handling: frozen task copies do not update automatically. The freezer reports missing cache metadata as `unknown_write_evict`; the director reads the receipt's `cache_mode` into `cache_basis`, flags legacy results, and prevents mixing measurement bases.

### Source bindings (validation from the preceding commits)

- Full Python test list from L0 CI, without coverage instrumentation: **2156 passed, 5 skipped, 90 subtests passed**.
- CPU C++ regression preserves the faulty skip-existing-link builder: the old copy misses a candidate `#error`; the fixed materializer detects it through Engineer, Verify and next-wave copies.
- Runtime tests cover stale links present before a command or generated during it, swallowed input-validation errors, legitimate shared references, ancestry across copies, and source-binding helpers.
- Candidate-floor/recovery and expert-skills Node regressions passed; shell syntax, dry-run handoff mapping and Ruff on both new Python files passed. Both regression suites are wired into L0.

### Fixed read eviction

- Current CPU regression run: **465 tests passed** across `test_harness_lib`, `test_op_bench`, and `test_leg_runner`.
- Tests check actual reads before each timed sample in eager and graph paths, no eviction inside graph capture, removal of the old switches, exact buffer reuse, positive sizes, allocation before warmup, stable size after an environment change, CPU fallback, and receipt preservation.
- `git diff --check` passed. Ruff found no new diagnostics relative to the preceding commit in the changed Python files (86 pre-existing findings).
- The GPU experiments above were run on MI355X with the three-arm implementation at `9d6cd7bb`, before removing mode selection. Two independent runs reproduced per-cell within 1%; all nine probe cells carried `timer=cuda_event`, `primed=True`, `host_ms`, and `cache_condition`.
- **GPU regression of the final single-policy change passed** on MI355X / PyTorch 2.9.1 / ROCm 7.2. Compared pinned `9d6cd7bb` with `HARNESS_CACHE_MODE=read-evict` against current `ddfd0cc8`, in two independent processes on the same idle GPU. Each workload/timer has 12 alternating old/current pairs per process, 10 warmups and 50 samples per timing call.

| Workload | Timer | Paired latency change, current vs old | 95% bootstrap interval |
|---|---|---:|---:|
| 64 MiB sum | eager | +0.01% | [-0.09%, +0.11%] |
| 64 MiB sum | graph replay | -0.08% | [-0.26%, +0.09%] |
| 4 GiB sum | eager | -0.14% | [-0.32%, +0.06%] |
| 4 GiB sum | graph replay | +0.20% | [-0.06%, +0.47%] |

The equivalence band was fixed at ±5% before execution; all intervals fit within ±0.5%. The independent copy control drifted -0.18% / +0.18% across the two runs. All 192 timing receipts had the requested event timer (no graph fallback), `primed=True`, `host_ms`, and the expected `cache_condition`; eviction buffers were reused and reduction results checked. Eight end-to-end `leg_runner` checks and four GPU checks of ignored legacy environment values also passed. This validates the harness simplification; it does not remeasure historical model/kernel speedups.

Existing checks: repository-wide Ruff reports 1055 advisory findings. The additional `test_mode_dispatch.js` run has two winner-selection assertion failures; both reproduce on unmodified main `0e02ad3a`.

No full generative GPU optimization run was performed. The runtime guard covers workspace source links; arbitrary compiler/load paths also rely on the builder's input check and relocation probe, rather than compiler interception. No recorded task was re-measured for the cache change — already-frozen tasks keep their numbers and self-identify as `unknown_write_evict` until they are re-frozen.

## What changes for a caller

`time_op(call, ..., detail=True)` now always uses read eviction on GPU. All in-repository production callers already use this form and need no changes.

- Remove any explicit `flush_cache` argument from `time_op` calls and any `mode` argument from direct `flush_cache` calls. These removed arguments raise `TypeError`; `detail` is keyword-only so an old positional flush switch cannot silently bind to it.
- Remove `HARNESS_CACHE_MODE` from launch environments. It is no longer read, so stale values cannot disable eviction or restore writes.
- `HARNESS_CACHE_FLUSH_MB` is the remaining cache setting: a positive integer number of MiB, default `512`. Invalid sizes raise before running the timed callable.

### Files

Fix 2 touches five files:

| file | change |
|---|---|
| `e2e_workflow/scripts/harness_lib.py` | fixed read eviction, exact-sized buffer, `cache_condition` on the receipt |
| `e2e_workflow/scripts/tests/test_harness_lib.py` | timing, eviction, receipt, removed-switch, and CPU fallback regressions |
| `e2e_workflow/scripts/leg_runner.py` | projects `cache_condition` into the per-leg timing dict |
| `e2e_workflow/scripts/op_bench.py` | `_time_call` no longer documents "cache flushed cold each sample" |
| `e2e_workflow/roles/op_benchmarker.md` | same wording fix in the role prompt, plus why pre-change decode numbers are on a different basis |

plus the `cache_basis` block in `kernel_workflow/roles/director.md` and the `cache_mode` receipt field in
`kernel_workflow/roles/oracle_freezer.md` — the two places that have to read the label.

## What this does to numbers already on the books

Nothing is re-measured here, and nothing self-heals: `harness_lib.py` reaches a task directory by `cp`
(`kernel_extractor.md`, `oracle_freezer.md`), so every task frozen before this merge keeps a write-evict
harness forever. The freshness gate already stops a *new* freeze from reusing a stale copy, but it does
not retroactively invalidate old scores — hence the label rather than a silent fix:

- A receipt **with** `cache_condition` is self-describing; read `mode`.
- A receipt **without** one was produced under the unconditional write-evict. `director.md` records
  `cache_basis: "unknown_write_evict"` and `status: "flagged"`. Absence is deliberately *not* read as
  "no cache preparation".
- There is **no correction factor**. The inflation scales with each leg's HBM traffic, so it moves
  different cases by different amounts and is not invertible from the recorded number. Re-freezing
  against a current `$HARNESS_LIB` is the only fix.
- Two different bases must never be folded into one campaign-level speedup.

Decode-served tasks are where this bites; prefill-only tasks are largely immune but still carry the label.

## Reproducing the cache measurements

The experiment artifacts live outside the repo:

- `moe_flush_probe_0921/FINDINGS.md` — the GLM-5.2 fused-MoE A/B across the three
  modes, with the null arm and per-bucket MDE that make the 1.40-vs-1.12 gap readable.
- `cache_mode_probe_0922/` — the absolute-time probe supporting read eviction:
  `PROBE.md` (design, data, reading), `probe.py` (three arms × three modes, using the experimental
  `harness_lib.time_op(detail=True)` at commit `9d6cd7bb`; mode selection is removed from the current code), `run.sh <label> <gpu>`, `probe_r{1,2}.jsonl` (per-cell raw data
  plus a `_meta` record with the contention bracket, device and harness path), and `legsmoke_task/` for
  the end-to-end `leg_runner` smoke.

The final-code regression is recorded separately in `cache_read_regression_20260922T090009Z/`:
`REPORT.md`, `summary.json`, `run_1.jsonl`, `run_2.jsonl`, the pinned source hashes in `manifest.json`,
and the runnable `probe.py`, `analyze.py`, and `run_job.sh`. It compares old read eviction with the
current fixed implementation; it does not restore mode selection in production code.

The probe's floor arm is measured, not assumed: a 64 KiB reduction is too small to be about memory, and
it still moved 7.2 → 9.6 µs between modes, so the per-sample launch+event floor is subtracted per mode.
That floor is itself slightly inflated by eviction (the op's own operand goes cold), which makes the
subtraction conservative in the direction that *understates* the effect reported above.

## Scope

Not in this PR, tracked separately: the two remaining `dsa_indexer` defects — zero coverage of the
`M = bs` EAGLE/MTP draft shape, and the absence of a standing gate for a candidate's own unbounded
shape-keyed launch cache. Neither is affected by the changes here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yueliu14:
GPU follow-up validation of commit `4cedb8eb` completed on MI355X / gfx950 with ROCm 7.2, Torch 2.9.1 and Triton 3.7.0. Tested four real, self-contained kernel tasks using their existing builders and UTs:

| Build path | Unmodified full UT | Direct candidate compile-error probe | Copied parent-link control | Fixed materialization + probe | Second Verify copy + probe | Reintroduced stale-link guard |
|---|---|---|---|---|---|---|
| AITER HIP RMSNorm | Pass | Detected | Wrong parent source compiled and executed | Detected | Detected | Exit 86 before command |
| SGLang HIP top-k | Pass | Detected | Wrong parent source compiled and executed | Detected | Detected | Exit 86 before command |
| Gluon indexer decode | Pass | Detected | Wrong parent source compiled and executed | Detected | Detected | Exit 86 before command |
| Triton sparse MLA prefill | Pass | Detected | Wrong parent source compiled and executed | Detected | Detected | Exit 86 before command |

All **24 valid scenarios** produced the expected outcome. The HIP probes use `#error`; the Triton/Gluon probes use compile-time `static_assert(False, ...)` in candidate JIT bodies. This verifies actual compilation, not merely Python import. Compiler/loader paths and SHA256 values were recorded: the retained-link controls read the original hash, while both fixed copy generations read the edited candidate hash. AITER Ninja rules and SGLang generated build files were also retained locally. Runtime-guard checks recorded invalid-measurement audits and did not execute the test command.

Scope: the four original tasks did not already contain stale source links. Each received a separate disposable fixture reproducing the parent-link mechanism, while preserving its real builder and include context. The positive controls used the unchanged full UT entry point (including all three cache modes where provided). This was not a fresh generative GEAK optimization run or a service e2e/performance-gain certification. Original task files remained byte-identical.

Five attempts were excluded and rerun: three stopped before kernel execution because the diagnostic logging wrapper initially referenced an incorrect SGLang internal export; two were refused by the GPU idle guard. The logging wrapper was corrected, and the rejected GPU attempts were repeated once the node was idle; the idle guard was not disabled. No GEAK implementation change was needed following these tests.

yueliu14:
Added a complete synthetic wrong-output regression in `e7563adb`, including a reusable task fixture and a CPU-only L0 test.

The independent oracle expects `y = 2*x`. The candidate is deliberately changed to `y = 3*x`; both versions compile and execute successfully. The harness retains the same faulty `_mirror()` skip-existing-link logic throughout.

| Scenario | Actual output for `[1, 2, -3, 0.5]` | Result |
|---|---|---|
| Correct parent | `[2, 4, -6, 1]` | Pass |
| Old materialization + bad candidate | `[2, 4, -6, 1]` from the parent | **False pass** |
| PR materialization + identical bad candidate | `[3, 6, -9, 1.5]` | Correctness fail, exit 1; ineligible for benchmarking |
| Second Verify copy | `[3, 6, -9, 1.5]` | Correctness fail, exit 1 |
| Repaired candidate | `[2, 4, -6, 1]` | Pass |
| Stale parent link reintroduced | Command not executed | Invalid measurement, exit 86 |

The standalone run compared the actual `materialize_workspace.sh` from main `0e02ad3a` against the PR version. The harness was byte-identical across all six scenarios, as were the bad candidate sources across old/fixed/Verify directories. Every invocation rebuilt the executable, so stale binaries cannot explain the false pass.

Validation: the six standalone scenarios matched the expected outcomes; **32 related pytest tests passed**, and Ruff passed for the changed Python files. The new regression is in the test file already executed by L0 CI. Fixture and instructions: `kernel_workflow/scripts/tests/fixtures/source_binding/`.

This synthetic task is separate from the unavailable original incident archive. This run used host C++ execution. Its HIP/gfx950 branch also compiled successfully, but was not executed: the existing GPU node was busy and a new allocation was rejected with `QOSMaxSubmitJobPerUserLimit`. No existing jobs were cancelled or idle checks disabled. The earlier four real-kernel GPU tests remain separate evidence.

zihaoanllm (approved):

chao-xu-spec (approved):

